PR #77, opened 2021-09-08
Added update function and corresponding tests for vmc_networks module

shaikres:
Added update function and corresponding tests for vmc_networks module

### DIFF
--- a/src/saltext/vmware/modules/vmc_networks.py
+++ b/src/saltext/vmware/modules/vmc_networks.py
@@ -238,3 +238,277 @@ def delete(
         verify_ssl=verify_ssl,
         cert=cert,
     )
+
+
+def update(
+    hostname,
+    refresh_key,
+    authorization_host,
+    org_id,
+    sddc_id,
+    network_id,
+    verify_ssl=True,
+    cert=None,
+    subnets=vmc_constants.VMC_NONE,
+    admin_state=None,
+    description=None,
+    domain_name=None,
+    tags=vmc_constants.VMC_NONE,
+    advanced_config=None,
+    l2_extension=vmc_constants.VMC_NONE,
+    dhcp_config_path=None,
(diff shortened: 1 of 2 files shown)


src/saltext/vmware/modules/vmc_networks.py
```diff
@@ -238,3 +238,277 @@ def delete(
...
+        base_url=api_url_base, org_id=org_id, sddc_id=sddc_id, network_id=network_id
+    )
+
+    existing_data = get_by_id(
+        hostname, refresh_key, authorization_host, org_id, sddc_id, network_id, verify_ssl, cert
+    )
+
+    # fetch the network for the given network_id
+    if vmc_constants.ERROR in existing_data:
+        return existing_data
```
waynew, line 481: **suggestion** A test for this existing_data is missing from this PR. Not sure if there's a test covering this on `main` already... the easy way to check is to change this to 

```
if vmc_constants.ERROR in existing_data:
    assert False
    return existing_data
```

Then run `python -m nox -e tests-3 -- tests/unit/`. Currently nothing fails, so tests aren't covering this case.

waynew (changes requested):
Looks like a missing test case, in addition to needing to rebase on main. Might be easiest to open a new PR? :thinking:

src/saltext/vmware/modules/vmc_networks.py
shaikres: tests/unit/modules/test_vmc_networks.py has already included testcase `test_update_network_when_api_returns_no_network_with_given_id` to cover the mentioned scenario

shaikres:
Closing this PR as it is having merge conflicts. Raised new PR [https://github.com/saltstack/salt-ext-modules-vmware/pull/107](url) for the same code.

PR #81, opened 2021-09-13
Added get function for vmc distributed firewall rules and corresponding unit tests

navalkp:


src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
```diff
@@ -0,0 +1,118 @@
...
+    api_url = api_url.format(
+        base_url=api_url_base,
+        org_id=org_id,
+        sddc_id=sddc_id,
+        domain_id=domain_id,
+        security_policy_id=security_policy_id,
+    )
+
+    params = vmc_request._filter_kwargs(
+        allowed_kwargs=["cursor", "page_size", "sort_ascending", "sort_by"],
```
waynew, line 103: **suggestion** looks like this PR is missing tests to ensure the params are correct. After this line I was able to replace params with `params = None` and all the tests continue to pass.

waynew (changes requested):
Missing some tests to ensure that correct params are passed. Otherwise this looks good :+1:

src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
jain-prerna: Hi @waynew , we agree with this suggestion but as we already done the implementation through out, would be difficult to address this for now. we can surely take this as further test improvement later.

PR #83, opened 2021-09-13
Added get method for vmc_vm_stats module and corresponding tests

shaikres:
Added get function for vmc_vm_stats module and corresponding tests
Added supported util functions

### DIFF
--- a/src/saltext/vmware/utils/vmc_vcenter_request.py
+++ b/src/saltext/vmware/utils/vmc_vcenter_request.py
@@ -1,0 +1,144 @@
+"""
+    VMC vCenter API Request Module
+"""
+import logging
+
+import requests
+from requests.auth import HTTPBasicAuth
+from requests.exceptions import HTTPError
+from requests.exceptions import RequestException
+from requests.exceptions import SSLError
+from saltext.vmware.utils import vmc_constants
+from saltext.vmware.utils import vmc_request
+
+log = logging.getLogger(__name__)
+
+
+def get_api_session_id(hostname, username, password):
+    """
+    This function returns api_session_id required to perform operations for VMC vCenter.
+
+    hostname
+        Hostname of the vCenter console
+
+    username
+        username required to login to vCenter console
+
+    password
+        password required to login to vCenter console
+
+    """
+    url = vmc_request.set_base_url(hostname) + vmc_constants.VCENTER_API_SESSION_URL
+    headers = {vmc_constants.CONTENT_TYPE: vmc_constants.APPLICATION_JSON}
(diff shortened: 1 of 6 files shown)


tests/unit/conftest.py
```diff
@@ -20,6 +26,14 @@ def mock_access_token(mock_response):
         yield vmc_access_token
 
 
+@pytest.fixture
+def mock_vcenter_headers(mock_response):
+    with patch("saltext.vmware.utils.vmc_vcenter_request.get_headers") as vcenter_headers:
+        mock_response.json = mock.Mock("mock-header")
+        vcenter_headers.return_value = mock_response
+        yield vcenter_headers
```
waynew, line 34: **suggestion** this fixture is probably less desirable. Currently it's possible to remove both `get_api_session_id` and `get_headers` from `utils/vmc_vcenter_request.py` and all the unit tests pass.

A possible approach would be just to mock `requests.post` to return something like:

```
with patch("requests.post", autospec=True) as fake_request:
    fake_request.json.return_value = {'expected json': 'dictionary goes here'}
    yield
```

src/saltext/vmware/utils/vmc_vcenter_request.py
```diff
@@ -0,0 +1,144 @@
...
+
+    """
+    api_session_id = get_api_session_id(hostname, username, password)
+    return {
+        vmc_constants.CONTENT_TYPE: vmc_constants.APPLICATION_JSON,
+        vmc_constants.VMWARE_API_SESSION_ID: api_session_id,
+    }
+
+
+def call_api(
```
waynew, line 58: **suggestion** this appears to be a near/exact duplicate to `vmc_request`'s `call_api`. Is there a good reason not to have a common function for calling the API?

waynew (changes requested):
the `vmc_vm_stats` part looks good, but utils & tests need some things addressed

src/saltext/vmware/utils/vmc_vcenter_request.py
shaikres: Headers part is different for VMC and vcenter APIs.  In order to keep the code simple and understandable, we kept it as separate.

tests/unit/conftest.py
shaikres: For vcenter APIs to be working properly, we need to pass api_session_id (based on vcenter_host, username and password) as header. Therefore, we should not remove `get_api_session_id` and `get_headers` from `utils/vmc_vcenter_request.py`

shaikres:
Addressed the review comments replacing this PR with [https://github.com/saltstack/salt-ext-modules-vmware/pull/116](url)

PR #85, opened 2021-09-16
Added state module for vmc nat rules and corresponding unit tests

navalkp:


src/saltext/vmware/states/vmc_nat_rules.py
```diff
@@ -0,0 +1,512 @@
...
+                sequence_number=sequence_number,
+                display_name=display_name,
+            )
+
+            if "error" in updated_nat_rule:
+                return vmc_state._create_state_response(
+                    name=name, comment=updated_nat_rule["error"], result=False
+                )
+
+            get_nat_rule_after_update = __salt__["vmc_nat_rules.get_by_id"](
```
waynew, line 326: **question** why get after update? Does `updated_nat_rule` not contain the new information?

src/saltext/vmware/states/vmc_nat_rules.py
```diff
@@ -0,0 +1,512 @@
...
+        verify_ssl=verify_ssl,
+        cert=cert,
+    )
+
+    existing_nat_rule = None
+
+    if "error" not in get_nat_rule:
+        log.info("Nat rule found with Id %s", nat_rule)
+        existing_nat_rule = get_nat_rule
+    elif NAT_RULE_NOT_FOUND_ERROR not in get_nat_rule["error"]:
```
waynew, line 271: **suggestion** this error handling is confusing and filled with double-negatives. It would be clearer written like:

```
if "error" in get_nat_rule:
    if NAT_RULE_NOT_FOUND_ERROR in get_nat_rule["error"]:
        existing_nat_rule = None
    else:
        return vmc_state._create_state_response(...)
else:
    existing_nat_rule = get_nat_rule
```

src/saltext/vmware/states/vmc_nat_rules.py
```diff
@@ -0,0 +1,512 @@
...
+                    name=name, comment=get_nat_rule_after_update["error"], result=False
+                )
+
+            return vmc_state._create_state_response(
+                name=name,
+                comment="Updated nat rule {}".format(nat_rule),
+                old_state=existing_nat_rule,
+                new_state=get_nat_rule_after_update,
+                result=True,
+            )
```
waynew, line 350: **suggestion** This behavior could be delegated to `_create_state_response`, since `error` in new_state appears to always trigger `result=False`. It would be possible to update the response to override it like so:

```
result = 'error' in new_state if new_state and result is None else result
```

Of course, a result of `None` *does* have meaning, so if there is a case where the above approach doesn't work, the response function signature could be written like:

```
MISSING = object()

def _create_state_response(..., result=MISSING):
    ...
    result = 'error' in new_state if new_state and result is MISSING else result
```

src/saltext/vmware/states/vmc_nat_rules.py
```diff
@@ -0,0 +1,512 @@
...
+        else:
+            log.info("All fields are same as existing nat rule %s", nat_rule)
+            return vmc_state._create_state_response(
+                name=name, comment="Nat rule exists already, no action to perform", result=True
+            )
+    else:
+        log.info("No nat rule found with Id %s", nat_rule)
+        created_nat_rule = __salt__["vmc_nat_rules.create"](
```
waynew, line 358: **suggestion** for many/most of these calls out to the modules, no tests are ensuring that they're being called correctly. Are these checked in one of the integration test PRs?

waynew: (note - `create_autospec` is at least ensuring that the required arguments are present, but it would be appropriate to at least test that all the arguments are actually being passed to these underlying mocks.)

waynew (changes requested):

src/saltext/vmware/states/vmc_nat_rules.py
```diff
@@ -0,0 +1,512 @@
...
+        "source_network": source_network,
+        "tags": tags,
+        "translated_network": translated_network,
+        "translated_ports": translated_ports,
+        "enabled": enabled,
+        "firewall_match": firewall_match,
+        "logging": logging,
+        "sequence_number": sequence_number,
+        "display_name": display_name,
+    }
```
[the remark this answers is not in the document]
navalkp: added new test for actual calls.

src/saltext/vmware/states/vmc_nat_rules.py
navalkp: I looked at it, we have comment field which changes based on failure and success. So keeping it as it is, making no change

waynew:
Closing PR in favor of #117

PR #86, opened 2021-09-16
Added integration tests for vmc_dhcp_profiles execution module

shaikres:
Added integration tests for vmc_dhcp_profiles execution module

### DIFF
--- a/tests/integration/modules/test_vmc_dhcp_profiles.py
+++ b/tests/integration/modules/test_vmc_dhcp_profiles.py
@@ -1,0 +1,193 @@
+"""
+    Integration Tests for vmc_dhcp_profiles execution module
+"""
+import pytest
+import requests
+from saltext.vmware.utils import vmc_constants
+from saltext.vmware.utils import vmc_request
+
+
+@pytest.fixture(scope="module")
+def dhcp_test_data():
+    dhcp_profile_type = "relay"
+    dhcp_profile_id = "dhcp-test"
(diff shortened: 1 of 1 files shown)


waynew:
@shaikres for the integration tests, how are we spinning up VMC/NSX? Might be a discussion for internal Slack :thinking:

waynew (changes requested):
**suggestion** while the way this is written *does* pass tests running locally, the way it's written is not very clear at all, and is very difficult to follow. 
We could review it during one of the [Test Clinics on Twitch](https://www.twitch.tv/saltstackinc/schedule), but generally there's a lot of boilerplate/duplication setup code.

There may be a better way, but my first impression is that these tests should basically look like this:

```
def test_create_dhcp_profile(salt_call_cli, vmc_dhcp_create_kwargs, dhcp_profile_id):
    ret = salt_call_cli.run("vmc_dhcp_profiles.create", **vmc_dhcp_create_kwargs)
    result_as_json = ret.json
    assert result_as_json["id"] == result_as_json["display_name"] == dhcp_profile_id
```

That might not be the best approach (maybe `server_addresses` should also be provided within the test, for instance?), but generally peeling the common args/functionality into a fixture or fixtures tends to make tests easier to follow.

waynew:
Closing because this PR is from an unlinked fork (thanks, GitHub!) - please open a new one that addresses this review

PR #87, opened 2021-09-16
Added integration tests for vmc nat rules execution module

navalkp:


waynew (changes requested):
**suggestion** On reviewing this I notice the same issue as #86 - it looks like there could be a better way to write these tests.

Another thing noticed when running all the vmc_nat tests:

    python -m nox -e tests-3 -- tests/ -k 'vmc_nat'

Shows this file has some missing coverage:

    src/saltext/vmware/modules/vmc_nat_rules.py                     53      2      2      1    95%   18, 745->746, 746
    
That corresponds to https://github.com/jain-prerna/salt-ext-modules-vmware-old/blob/e517a54c715ee84622fc200c2d29e9d66a5d53e5/src/saltext/vmware/modules/vmc_nat_rules.py#L745-L746

It should be relatively simple to add a unit test for that as part of the (new) PR

I just noticed that these PRs are on the broken fork, so I'm going to go ahead and close #86 as well as #87 - go ahead and open new PRs from these branches in

PR #96, opened 2021-09-30
Manage hosts on switch.

cheburakshu:
#SSC-101

### DIFF
--- a/src/saltext/vmware/modules/dvswitch.py
+++ b/src/saltext/vmware/modules/dvswitch.py
@@ -1,0 +1,283 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import logging
+
+import salt.exceptions
+import saltext.vmware.utils.common as utils_common
+import saltext.vmware.utils.datacenter as utils_datacenter
+import saltext.vmware.utils.vmware as utils_vmware
+from saltext.vmware.utils.connect import get_service_instance
+
+log = logging.getLogger(__name__)
+
+try:
+    from pyVmomi import vim, vmodl
+
+    HAS_PYVMOMI = True
+except ImportError:
+    HAS_PYVMOMI = False
+
+__virtualname__ = "vmware_dvswitch"
+__proxyenabled__ = ["vmware_dvswitch"]
+
+
+def __virtual__():
+    if not HAS_PYVMOMI:
+        return False, "Unable to import pyVmomi module."
+    return __virtualname__
+
+
+def _get_switch_config_spec(service_instance, datacenter_name, switch_name):
+    dc_ref = switch_ref = config_spec = None
+    dc_ref = utils_datacenter.get_datacenter(service_instance, datacenter_name)
+    switch_refs = utils_vmware.get_dvss(dc_ref=dc_ref, dvs_names=[switch_name])
+    if switch_refs:
+        switch_ref = switch_refs[0]
+        dvs_props = utils_common.get_properties_of_managed_object(
+            switch_ref, ["config", "capability"]
+        )
+        config_spec = vim.dvs.VmwareDistributedVirtualSwitch.ConfigSpec()
+        # Copy all of the properties in the config of the of the DVS to a
+        # DvsConfigSpec
+        skipped_properties = ["host"]
+        for prop in config_spec.__dict__.keys():
+            if prop in skipped_properties:
+                continue
+            if hasattr(dvs_props["config"], prop):
+                setattr(config_spec, prop, getattr(dvs_props["config"], prop))
+    return dc_ref, switch_ref, config_spec
+
+
+def manage(
+    datacenter_name,
+    switch_name,
+    uplink_count=None,
+    uplink_prefix="Uplink ",
+    switch_version=None,
+    switch_description=None,
+    mtu=None,
+    discovery_protocol=None,
+    discovery_operation=None,
+    multicast_filtering_mode=None,
+    contact_name=None,
+    contact_description=None,
+    network_forged_transmits=None,
+    network_mac_changes=None,
+    network_promiscuous=None,
+    health_check_teaming_failover=None,
+    health_teaming_failover_interval=None,
+    health_vlan_mtu=None,
+    health_vlan_mtu_interval=None,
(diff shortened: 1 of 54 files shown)


src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,155 @@ def manage(
...
+    host_name=None,
+    datacenter_name=None,
+    cluster_name=None,
+    nics=None,
+    num_ports=None,
+    mtu=None,
+    service_instance=None,
+):
+    """
+    Add/update an ESXi host to a distributed vSwitch.
```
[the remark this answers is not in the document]
dhiltonp: I guess there's also `manage` above, we could change that for consistency...

I don't think `manage` is a function anywhere else in the project

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,155 @@ def manage(
...
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    nics
+        List of vmnics to attach to vSwtich. (optional). Default "None".
+
+    num_ports
+        Number of ports to configure on the vSwitch. (optional). Default 128.
```
dhiltonp, line 375: I'm confused by this - we're not creating the switch, and when you set up a switch (like a 24 port switch or whatever), usually only 1 port would go to the switch from a host...

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,155 @@ def manage(
...
+
+    mtu
+        MTU to configure on the vSwitch. (optional). Default 1600.
```
dhiltonp, line 378: we can reconfigure the MTU? I suspect we're configuring a virtual NIC's MTU, not the vSwitch's MTU?

src/saltext/vmware/modules/dvswitch.py
cheburakshu: The api lets the user configure/re-configure the params on the vswitch itself.

```
                    vss_spec.mtu = mtu or switch.mtu
                    vss_spec.numPorts = num_ports or switch.spec.numPorts
```

src/saltext/vmware/modules/dvswitch.py
cheburakshu: same as above

src/saltext/vmware/modules/dvswitch.py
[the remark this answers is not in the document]
cheburakshu: I would change to add or update host, as it really doesn't configure a host.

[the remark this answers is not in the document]
dhiltonp: that's true, but it is also within the context of dvswitch...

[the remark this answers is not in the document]
dhiltonp: oh, and also this can manage multiple hosts? like add all of the hosts in a cluster to a dvswitch?

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,155 @@ def manage(
...
+    )
+    try:
+        for h in hosts:
+            ret[h.name] = False
+            network_manager = h.configManager.networkSystem
+            if not network_manager:
+                continue
+            if isinstance(nics, str):
+                nics = [nics]
+            vss_spec = vim.host.VirtualSwitch.Specification()
```
dhiltonp, line 407: VirtualSwitch is inside of host? This has to do something different from vim.dvs.VmwareDistributedVirtualSwitch... [I've found the docs for the latter](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/9fd87c06-14a3-41e5-b28d-277864a80f29/d6112c2a-b124-4fa5-96d7-9fb4b6f1bb50/doc/vim.dvs.VmwareDistributedVirtualSwitch.html), but I'm having a little difficulty finding the docs for the former (vim.host.VirtualSwitch).

I'm really trying to understand what is going on here, because it seems so weird in the context of what I know about networking.

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,155 @@ def manage(
...
+            if hasattr(network_manager.networkInfo, "vswitch"):
```
dhiltonp, line 408: what if this is false? Nothing happens?

dhiltonp: oh, got it, `ret[h.name]` is the else... I understand how that's sometimes the easiest way to say that, it's just not always the most obvious :/

dhiltonp: maybe renaming `ret` to `updated` would improve readability?

src/saltext/vmware/modules/dvswitch.py
cheburakshu: good point, yes. it is a standard switch. wondering if I should move it to a vswitch module instead of this dvswitch.

src/saltext/vmware/modules/dvswitch.py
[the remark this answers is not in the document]
cheburakshu: yes, it can.

src/saltext/vmware/modules/dvswitch.py
cheburakshu: I don't have a sure way to do it. the switch_name here is a dvswitch and it lets the user configure params on the vswitch on the host, not the dvswitch param itself. So, I am going to leave it there in the dvswitch module.

cheburakshu:
closing this PR and creating a new one

src/saltext/vmware/modules/dvswitch.py
dhiltonp: ok, that makes some things more clear

the dvswitch is a switch between regular virtual switches, this function creates a regular vswitch on the host, that is connected to the dvswitch

dhiltonp: right?

dhiltonp:
#102

PR #101, opened 2021-10-01
Furo theme

ScriptAutomate:
Update to use Furo theme


waynew:
Looks like there was a leftover import. Went ahead and fixed/pushed, should be good to merge after checks :+1:

waynew (approved):

PR #102, opened 2021-10-01
Manage host on vswitch

cheburakshu:
#ssc-101


src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,228 @@ def manage(
...
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    nics
+        List of vmnics to attach to vSwtich. (optional). Default "None".
```
dhiltonp, line 372: typo

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,228 @@ def manage(
...
+    host_name=None,
+    datacenter_name=None,
+    cluster_name=None,
+    nics=None,
+    num_ports=128,
+    mtu=1500,
+    service_instance=None,
+):
+    """
+    Add an ESXi host to a distributed vSwitch.
```
dhiltonp, line 357: so this creates a regular vSwitch exclusive to the ESXi host and connects it to the distributed vSwitch passed in?

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,228 @@ def manage(
...
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    nics
+        List of vmnics to attach to vSwtich. (optional). Default "None".
```
dhiltonp, line 452: typo in vSwitch

src/saltext/vmware/modules/dvswitch.py
dhiltonp: if so, I think that updating the docstring to reflect what is happening makes sense

dhiltonp: Something like `Create a new vSwitch on the host linked to the distributed switch in switch_name.`

dhiltonp: unless I still don't understand what's going on

cheburakshu: `Add new hosts to this distributed switch.`, `Manage networking of hosts attached to this distributed switch.`, `Remove hosts from this distributed switch.` are the descriptions available in the UI. I am not sure if it is correct to add a description like `Create a new vSwitch on the host`, as I don't know if vCenter really creates one, though it seems it does. There is no documentation to that effect.

src/saltext/vmware/modules/dvswitch.py
cheburakshu: will change

src/saltext/vmware/modules/dvswitch.py
cheburakshu: will change

src/saltext/vmware/modules/dvswitch.py
dhiltonp: hmmm, ok. I'm really trying to understand how mtu and ports come into play, because they aren't affecting the ports on the vSwitch.

cheburakshu: seems it alters the mtu and the ports on the switch on the host. Don't see such an option in the UI, but the api does give it. I updated the mtu, it doesn't change the distributed vswitch mtu, but updates the vswitch object's mtu on the host (printed on the console). Not a way to see it in the UI, at least I don't know where to see it.

dhiltonp: I wonder if when you use the UI it makes sure the host's switch isn't full, if it is it silently modifies the host's vswitch, adding another port if it's already full?

cheburakshu: it makes one regardless on the host.

cheburakshu: just printed out a configuration after adding a host manually

dhiltonp: That is interesting. Should we do the same thing and increment the number of ports when `add_host` is called?

Exposing num_ports and mtu on this call is confusing.

cheburakshu: We should only set it to the value, even if it is wrong and not increment as there isn't a way to decrement. Ansible does the same in their module too.

dhiltonp: I'm trying to find someone who works on vsphere to chime in, I did find references to `AddVirtualSwitch` failing due to stale vswitches which is interesting.

cheburakshu: @dhiltonp I am going to grab the mtu and num_ports from the dvswitch and set it on the spec, instead of allowing the user to specify some arbitrary mtu and num_ports. I am not finding anywhere in the UI where it lets the user set both these params at the host level. These can only be set at the switch level.

dhiltonp: I like that, if the user needs to set the properties of the switch then they can do a separate call.

src/saltext/vmware/modules/dvswitch.py
```diff
@@ -281,3 +282,232 @@ def manage(
...
+        vim.fault.NotFound,
+        vim.fault.HostConfigFault,
+        vim.fault.ResourceInUse,
+        vmodl.fault.InvalidArgument,
+        salt.exceptions.VMwareApiError,
+    ) as exc:
+        raise salt.exceptions.SaltException(str(exc))
+
+
+def add_host(
```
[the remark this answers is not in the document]
cheburakshu: ok will change

dhiltonp (approved):

PR #109, opened 2021-10-05
Manage esxi host

joechainz:


src/saltext/vmware/modules/esxi.py
```diff
@@ -732,3 +732,114 @@ def set_advanced_config(
         host_name=host_name,
         service_instance=service_instance,
     )
+
+
+def connect(name, service_instance=None):
+    """
+    Connect an esxi instance to a vCenter instance.
```
waynew, line 739: **suggestion (non-blocking)**  `ESXi` is the correct spelling

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,164 @@ def list_hosts(service_instance):
...
+    """
+    host_ref = utils_common.get_mor_by_property(service_instance, vim.HostSystem, name)
+    cluster_ref = utils_common.get_mor_by_property(
+        service_instance, vim.ClusterComputeResource, cluster_name
+    )
+    if cluster_ref.parent.parent != host_ref.parent.parent.parent:
+        raise salt.exceptions.VMwareApiError("Cluster has to be in the same datacenter")
+    task = cluster_ref.MoveInto_Task([host_ref])
+    utils_common.wait_for_task(task, cluster_name, "move host task")
+    return "moved"
```
waynew, line 227: **suggestion (non-blocking)** Would it be better to say where it moved from and where it moved to? e.g. `return f"moved {name} from {host_ref.parent.or.whatever} to {cluster_ref}"`?

Also, that raises a question - will it always be the `.parent.parent` and `.parent.parent.parent` of these things that is the datacenter, or is it possible that more or less levels of nesting could happen? Would it be better to do some kind of search up the tree for the datacenter?

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,164 @@ def list_hosts(service_instance):
...
+    name
+        Name of host.
+
+    service_instance
+        The Service Instance Object from which to obtain host.
+    """
+    host = utils_common.get_mor_by_property(service_instance, vim.HostSystem, name)
+    task = host.Destroy_Task()
+    utils_common.wait_for_task(task, name, "destroy host task")
+    return "removed"
```
waynew, line 245: **question** should we also note the name here?

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,164 @@ def list_hosts(service_instance):
...
+        IP address of host.
+    """
+
+    ctx = ssl.SSLContext()
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as _socket:
+        _socket.settimeout(1)
+        with ctx.wrap_socket(_socket) as wrappedSocket:
+
+            try:
+                wrappedSocket.connect((ip, 443))
```
waynew, line 273: **question** will this *always* use port 443, or is it configurable on the vCenter side?

waynew (changes requested):
Couple of feedback items - also just to double check code coverage-wise, I notice the integration tests are pretty light - do they actually cover all of the branches here?

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,164 @@ def list_hosts(service_instance):
...
+
+def _get_host_thumbprint(ip):
+    """
+    Returns host's ssl thumbprint.
+
+    ip
+        IP address of host.
+    """
+
+    ctx = ssl.SSLContext()
```
dhiltonp, line 267: does `with ssl.SSLContext() as ctx:` not work? I'm a little confused.

dhiltonp: That's pretty odd, I guess its naming predates python context managers... from python 2.5? Nope, ssl was introduced in 2.6 :shrug:

https://docs.python.org/3/library/ssl.html#functions-constants-and-exceptions

Should we use `context = ssl.create_default_context()` or `_create_unverified_context()`?

They do a bunch of configuration of the context, I think they're worth looking at:

https://github.com/python/cpython/blob/main/Lib/ssl.py#L745

https://github.com/python/cpython/blob/main/Lib/ssl.py#L783

dhiltonp: https://docs.python.org/3/library/ssl.html#security-considerations

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,167 @@ def list_hosts(service_instance):
...
+
+def _get_host_thumbprint(ip):
+    """
+    Returns host's ssl thumbprint.
+
+    ip
+        IP address of host.
+    """
+
+    ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
```
dhiltonp, line 270: I'm sorry for saying this now.

It's common to have unverified certs in dev, we should probably pass a flag to `add_host` that can disable it, maybe `verify_certificate` defaulting to true?

src/saltext/vmware/modules/esxi.py
```diff
@@ -732,3 +732,131 @@ def set_advanced_config(
...
+    return {"state": state}
+
+
+def add(
+    ip,
+    root_user,
+    password,
+    cluster_name,
+    datacenter_name,
+    server_auth=True,
```
dhiltonp, line 818: I'm not sure I like `server_auth` as a name and its description in the body.

The fingerprint auth always happens, what we're verifying is that the hostname matches its certificate, which is signed by a certificate authority (I think?).

src/saltext/vmware/utils/esxi.py
```diff
@@ -155,3 +159,167 @@ def list_hosts(service_instance):
...
+
+def add_host(
+    ip, root_user, password, cluster_name, datacenter_name, server_auth, connect, service_instance
+):
+    """
+    Adds host from vCenter instance
+
+    Returns connection state of host
+
+    ip
```
dhiltonp, line 290: https://www.vmware.com/support/developer/vc-sdk/visdk400pubs/ReferenceGuide/vim.ClusterComputeResource.html#addHost

Maybe we should change this to be `host`, then specify it's either the ip address or a hostname that can be resolved by DNS on the cluster.

src/saltext/vmware/modules/esxi.py
waynew: Something had bit nagging at the back of my mind and I realize what it was - do we need to do this kind of validation ourselves or is it something that we can delegate to, say, pyvmomi or requests or something? :thinking:

PR #114, opened 2021-10-07
get dns config for esxi hosts

cheburakshu:
#ssc-119

### DIFF
--- a/tests/integration/modules/test_esxi.py
+++ b/tests/integration/modules/test_esxi.py
@@ -275,6 +275,7 @@ def test_get_dns_config(service_instance):
         datacenter_name="Datacenter",
         cluster_name="Cluster",
     )
+    assert ret
     for host in ret:
         assert ret[host]["ip"]
         assert ret[host]["host_name"]


src/saltext/vmware/modules/esxi.py
```diff
@@ -732,3 +732,62 @@ def set_advanced_config(
...
+    log.debug("Running vmware_esxi.get_dns_config")
+    ret = {}
+    if not service_instance:
+        service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
+    hosts = utils_esxi.get_hosts(
+        service_instance=service_instance,
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
+        get_all_hosts=True if not host_name else False,
```
waynew, line 771: **question** would `bool(host_name)` read better here?

src/saltext/vmware/modules/esxi.py
```diff
@@ -732,3 +732,62 @@ def set_advanced_config(
...
+    )
+
+    try:
+        for h in hosts:
+            dns_config = h.config.network.dnsConfig
+            ret[h.name] = {}
```
waynew, line 777: **question** is it intentional that there will be an empty dict even when `not dns_config`? in other words, should this come after the `if not dns_config` block?

tests/integration/modules/test_esxi.py
```diff
@@ -264,3 +264,18 @@ def test_advanced_config(service_instance):
...
+def test_get_dns_config(service_instance):
+    """
+    Test get dns configuration on ESXi host
+    """
+    ret = esxi.get_dns_config(
+        service_instance=service_instance,
+        datacenter_name="Datacenter",
+        cluster_name="Cluster",
+    )
+    for host in ret:
```
waynew, line 278: **suggestion** note that this will pass without doing anything if there is nothing in `ret`, which would happen if `utils_esxi.get_hosts` returns empty. We should have either a unit test to confirm the behavior always (for instance, mocking get_hosts to return something expected/unexpected and ensure that the expected results are returned or the expected exception handling happens), or this test should probably fail when there are no results (since testing nothing is a useless test)

waynew (changes requested):
Mostly looks good - couple of minor items

src/saltext/vmware/modules/esxi.py
cheburakshu: done

tests/integration/modules/test_esxi.py
cheburakshu: done

waynew: :thinking: I'm not sure if I'm following - I see that the other test is checking when `ret` is *supposed* to be empty, but what about the first case when we expect it to be something?

Would 

```
assert ret
for host in ret:
 ...
 ```
 
 Be appropriate/correct?

cheburakshu:
> Other changes look good, just not sure about the test change

I added one more test to handle the previous comment when the function returns an empty dict when no host is matched. When one or more hosts are matched, it will return a dictionary of dns info indexed by host name.

> note that this will pass without doing anything if there is nothing in ret

PR #115, opened 2021-10-08
Added_function_to_manage_SDDC_host

jain-prerna:


src/saltext/vmware/modules/vmc_sddc_host.py
```diff
@@ -0,0 +1,201 @@
...
+    params = vmc_request._filter_kwargs(allowed_kwargs=["action"], action=action)
+    request_data = vmc_request.create_payload_for_request(vmc_templates.manage_sddc_host, req_data)
+    return vmc_request.call_api(
+        method=vmc_constants.POST_REQUEST_METHOD,
+        url=api_url,
+        refresh_key=refresh_key,
+        authorization_host=authorization_host,
+        description="vmc_sddc_host.manage",
+        data=request_data,
+        params=params,
```
waynew, line 144: **suggestion** these parameters are not being tested (i.e. they can be replaced with `None` and all tests still pass)

waynew: What I mean is that 

```
params = vmc_request._filter_kwargs(allowed_kwargs=["action"], action=action)
```

Could be replaced with

```
params = None
```

And there are no tests that would detect this problem.

jain-prerna: Added the required test.

jain-prerna:
> Looks like some tests are missing here, otherwise I didn't notice anything that needs to be addressed

I have added it, please do check now.

waynew (approved):
:+1:

PR #122, opened 2021-10-13
Get cluster info

cheburakshu:
#vrae-3246


src/saltext/vmware/modules/cluster.py
```diff
@@ -83,40 +83,48 @@ def create(name, datacenter, service_instance=None):
     return {name: True}
 
 
-def get_(name, datacenter, service_instance=None):
+def get_(cluster_name, datacenter_name, service_instance=None):
```
dhiltonp, line 86: Why is this `get_`? `list` has a name collision with python, so we need to use the func alias, but that shouldn't be the case for get...

src/saltext/vmware/modules/cluster.py
```diff
@@ -83,40 +83,48 @@ def create(name, datacenter, service_instance=None):
...
-
-    name
+    cluster_name
         The cluster name
 
-    datacenter
+    datacenter_name
         The datacenter name to which the cluster belongs
 
     .. code-block:: bash
```
dhiltonp, line 96: Should we have a docstring for service_instance?

cheburakshu: will add

src/saltext/vmware/modules/cluster.py
```diff
@@ -83,40 +83,48 @@ def create(name, datacenter, service_instance=None):
...
         service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
     try:
-        dc_ref = utils_datacenter.get_datacenter(service_instance, datacenter)
-        cluster_ref = utils_cluster.get_cluster(dc_ref=dc_ref, cluster=name)
+        dc_ref = utils_datacenter.get_datacenter(service_instance, datacenter_name)
+        cluster_ref = utils_cluster.get_cluster(dc_ref=dc_ref, cluster=cluster_name)
 
         # DRS config
         ret["drs_enabled"] = cluster_ref.configurationEx.drsConfig.enabled
+        ret["drs"] = __salt__["vmware_cluster_drs.get"](
```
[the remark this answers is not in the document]
cheburakshu: ok, will check

PR #123, opened 2021-10-13
List pkgs on esxi host

cheburakshu:


src/saltext/vmware/modules/esxi.py
```diff
@@ -919,3 +919,74 @@ def add(
...
+
+def list_pkgs(
+    pkg_name=None,
+    datacenter_name=None,
+    cluster_name=None,
+    host_name=None,
+    service_instance=None,
+):
+    """
+    List the packages installed on matching EXSi hosts.
```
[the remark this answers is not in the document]
dhiltonp: I'm sure it'll be convenient for small-medium size deployments, so I'm not advocating we remove the capability.

src/saltext/vmware/modules/esxi.py
```diff
@@ -919,3 +919,74 @@ def add(
...
+        for h in hosts:
+            host_pkg_manager = h.configManager.imageConfigManager
+            ret[h.name] = {}
+            if not host_pkg_manager:
+                continue
+            pkgs = host_pkg_manager.FetchSoftwarePackages()
+            for pkg in pkgs:
+                if pkg_name and pkg.name != pkg_name:
+                    continue
+                ret[h.name][pkg.name] = dict(
```
dhiltonp, line 975: I'm a little surprised by `dict(` vs `{`, not that it's wrong or anything.

src/saltext/vmware/modules/esxi.py
```diff
@@ -919,3 +919,74 @@ def add(
...
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
+        get_all_hosts=host_name is None,
+    )
+
+    try:
+        for h in hosts:
+            host_pkg_manager = h.configManager.imageConfigManager
+            ret[h.name] = {}
```
dhiltonp, line 968: I would swap these 2 lines to keep the host_pkg_manager lines together.

But it's really a tiny, tiny thing.

tests/integration/modules/test_esxi.py
```diff
@@ -124,6 +124,20 @@ def test_esxi_host_capability_params(service_instance, integration_test_config,
         assert capabilities[host_id][arg_name] == expected_value
 
 
+def test_list_pkgs(service_instance):
```
dhiltonp, line 127: The test seems a little weak - we verify that nothing raises and that there is some data within host?

I'm assuming that at least one package is always installed on a host by default, how about we verify that it contains data for the 7 keys we load?

src/saltext/vmware/modules/esxi.py
cheburakshu: can change.

src/saltext/vmware/modules/esxi.py
```diff
@@ -919,3 +919,74 @@ def add(
...
+                )
+        return ret
+    except (
+        vim.fault.InvalidState,
+        vim.fault.NotFound,
+        vim.fault.HostConfigFault,
+        vmodl.fault.InvalidArgument,
+        salt.exceptions.VMwareApiError,
+    ) as exc:
+        raise salt.exceptions.SaltException(str(exc))
```
dhiltonp, line 992: Do the errors from vCenter specific enough to not need additional context?

For example, do errors like NotFound report the host that doesn't exist?

src/saltext/vmware/modules/esxi.py
dhiltonp: Is there value in letting the user know that this host has no packages? Does that even happen?

src/saltext/vmware/modules/esxi.py
cheburakshu: yeah, they have a detailed message attached to them too.

dhiltonp (approved):

PR #142, opened 2021-11-16
Added integration tests for vmc_dhcp_profiles execution module. Update of PR-86

shaikres:
Added integration tests for vmc_dhcp_profiles execution module. Update of 
https://github.com/saltstack/salt-ext-modules-vmware/pull/86

### DIFF
--- a/tests/integration/modules/test_vmc_dhcp_profiles.py
+++ b/tests/integration/modules/test_vmc_dhcp_profiles.py
@@ -1,0 +1,193 @@
+"""
+    Integration Tests for vmc_dhcp_profiles execution module
+"""
+import pytest
+import requests
+from saltext.vmware.utils import vmc_constants
+from saltext.vmware.utils import vmc_request
+
+
+@pytest.fixture(scope="module")
+def dhcp_test_data():
+    dhcp_profile_type = "relay"
+    dhcp_profile_id = "dhcp-test"
+    return dhcp_profile_type, dhcp_profile_id
+
+
+@pytest.fixture
+def get_dhcp_profiles(vmc_nsx_connect, dhcp_test_data):
+    hostname, refresh_key, authorization_host, org_id, sddc_id, verify_ssl, cert = vmc_nsx_connect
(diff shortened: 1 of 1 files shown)


waynew (changes requested):
**suggestion** while the way this is written *does* pass tests running locally, the way it's written is not very clear at all, and is very difficult to follow. 
We could review it during one of the [Test Clinics on Twitch](https://www.twitch.tv/saltstackinc/schedule), but generally there's a lot of boilerplate/duplication setup code.

There may be a better way, but my first impression is that these tests should basically look like this:

```
def test_create_dhcp_profile(salt_call_cli, vmc_dhcp_create_kwargs, dhcp_profile_id):
    ret = salt_call_cli.run("vmc_dhcp_profiles.create", **vmc_dhcp_create_kwargs)
    result_as_json = ret.json
    assert result_as_json["id"] == result_as_json["display_name"] == dhcp_profile_id
```

That might not be the best approach (maybe `server_addresses` should also be provided within the test, for instance?), but generally peeling the common args/functionality into a fixture or fixtures tends to make tests easier to follow.

waynew:
@shaikres On today's Test Clinic I went through and refactored this set of integration tests into something that's more like what we want to see: [Like this](https://github.com/jain-prerna/salt-ext-modules-vmware-1/compare/vmc_dhcp_integration_tests...waynew:vmc_dhcp_integration_tests).

Let me know if you have any questions -- but that's the kind of setup that we're looking for :+1:

shaikres:
> @shaikres On today's Test Clinic I went through and refactored this set of integration tests into something that's more like what we want to see: [Like this](https://github.com/jain-prerna/salt-ext-modules-vmware-1/compare/vmc_dhcp_integration_tests...waynew:vmc_dhcp_integration_tests).
> 
> Let me know if you have any questions -- but that's the kind of setup that we're looking for 👍

Thanks Wayne, we have gone through the code and applied refactoring to our tests.

waynew (approved):

PR #146, opened 2021-11-18
vmc nat rules module it tests update of PR87

navalkp:
Added integration tests for vmc nat rules.
Added one unit test for missing test.


waynew (approved):
:+1:

PR #147, opened 2021-11-18
Get ntp config

cheburakshu:
ref: #vrae-3229

### DIFF
--- a/src/saltext/vmware/modules/esxi.py
+++ b/src/saltext/vmware/modules/esxi.py
@@ -783,6 +783,70 @@ def get_dns_config(
     return ret
 
 
+def get_ntp_config(
+    datacenter_name=None,
+    cluster_name=None,
+    host_name=None,
+    service_instance=None,
+):
+    """
+    Get NTP configuration on matching EXSI hosts.
+
+    datacenter_name
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    service_instance
+        Use this vCenter service connection instance instead of creating a new one. (optional).
+
+    .. code-block:: bash
+
+        salt '*' vmware_esxi.get_ntp_config
+    """
+    log.debug("Running vmware_esxi.get_ntp_config")
+    ret = {}
+    if not service_instance:
+        service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
+    hosts = utils_esxi.get_hosts(
+        service_instance=service_instance,
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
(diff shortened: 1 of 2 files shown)


src/saltext/vmware/modules/esxi.py
```diff
@@ -704,6 +704,70 @@ def get_dns_config(
...
+        datacenter_name=datacenter_name,
+        get_all_hosts=host_name is None,
+    )
+
+    try:
+        for h in hosts:
+            ntp_config = h.configManager.dateTimeSystem
+            if not ntp_config:
+                continue
+            ret[h.name] = {
```
waynew, line 749: **suggestion** could flip this:

```
if ntp_config:
    ret[h.name] = {...}
```

src/saltext/vmware/modules/esxi.py
```diff
@@ -704,6 +704,70 @@ def get_dns_config(
...
+                "time_zone": ntp_config.dateTimeInfo.timeZone.key,
+                "time_zone_name": ntp_config.dateTimeInfo.timeZone.name,
+                "time_zone_description": ntp_config.dateTimeInfo.timeZone.description,
+                "time_zone_gmt_offset": ntp_config.dateTimeInfo.timeZone.gmtOffset,
+                "ntp_servers": list(ntp_config.dateTimeInfo.ntpConfig.server),
+                "ntp_config_file": list(ntp_config.dateTimeInfo.ntpConfig.configFile)
+                if ntp_config.dateTimeInfo.ntpConfig.configFile
+                else None,
+            }
+        return ret
```
waynew, line 759: **suggestion** this early return is unnecessary either with or without changing the previous block.

tests/integration/modules/test_esxi.py
```diff
@@ -457,3 +457,34 @@ def test_esxi_get(service_instance):
...
+        assert sorted(list(ret[host])) == sorted(
+            [
+                "ntp_config_file",
+                "time_zone",
+                "time_zone_description",
+                "time_zone_name",
+                "ntp_servers",
+                "time_zone_gmt_offset",
+            ]
+        )
```
waynew, line 482: **question** can you use the test_value_scraper to grab the expected values, and then assert that the expected == actual? When we shift to actually configuring the ntp values on the hosts for the tests, this test wouldn't really have to be refactored at that point.

waynew (changes requested):
Looks good - just some minor changes here

tests/integration/modules/test_esxi.py
cheburakshu: ok

src/saltext/vmware/modules/esxi.py
cheburakshu: ok

src/saltext/vmware/modules/esxi.py
cheburakshu: ok

waynew (approved):
:+1:

PR #148, opened 2021-11-22
Vmc security groups module it test

navalkp:


waynew (changes requested):
``test_create_network_smoke_test`` fails with this error:
```
result_as_json = {'error': 'An object with the same path=[/infra/tier-1s/cgw/segments/web-tier] is marked for deletion. Either use another path or wait for the purge cycle (max 5 minutes) for permanen$
 removal of the object.'}
 ```
 
 Looks like the underlying code isn't handling that very well or the test should be doing something else.

waynew (approved):
Tests are passing now :+1:

PR #179, opened 2021-12-03
Manage users on ESXi hosts

cheburakshu:
ref: vrae-3235

### DIFF
--- a/src/saltext/vmware/modules/esxi.py
+++ b/src/saltext/vmware/modules/esxi.py
@@ -1093,7 +1093,7 @@ def remove_user(
 
 
 def get_user(
-    user_name=None,
+    user_name,
     datacenter_name=None,
     cluster_name=None,
     host_name=None,
--- a/src/saltext/vmware/modules/esxi.py
+++ b/src/saltext/vmware/modules/esxi.py
@@ -845,6 +845,326 @@ def get_ntp_config(
         raise salt.exceptions.SaltException(str(exc))
 
 
+def list_hosts(
+    datacenter_name=None,
+    cluster_name=None,
+    host_name=None,
+    service_instance=None,
+):
+    """
+    List ESXi hosts.
+
+    datacenter_name
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    service_instance
+        Use this vCenter service connection instance instead of creating a new one. (optional).
+
+    .. code-block:: bash
+
+        salt '*' vmware_esxi.list_hosts
+    """
+    log.debug("Running vmware_esxi.list_hosts")
+    ret = []
+    if not service_instance:
+        service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
+    hosts = utils_esxi.get_hosts(
+        service_instance=service_instance,
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
+        get_all_hosts=host_name is None,
+    )
+
+    try:
+        for h in hosts:
+            ret.append(h.name)
+        return ret
+    except (
+        vim.fault.InvalidState,
+        vim.fault.NotFound,
+        vim.fault.HostConfigFault,
+        vmodl.fault.InvalidArgument,
+        salt.exceptions.VMwareApiError,
+        vim.fault.AlreadyExists,
+    ) as exc:
+        raise salt.exceptions.SaltException(str(exc))
+
+
+def add_user(
+    user_name,
+    password,
+    description=None,
+    datacenter_name=None,
+    cluster_name=None,
+    host_name=None,
+    service_instance=None,
+):
+    """
+    Add local user on matching ESXi hosts.
+
+    user_name
+        User to create on matching ESXi hosts. (required).
+
+    password
+        Password for the new user. (required).
+
+    description
+        Description for the new user. (optional).
+
+    datacenter_name
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    service_instance
+        Use this vCenter service connection instance instead of creating a new one. (optional).
+
+    .. code-block:: bash
+
+        salt '*' vmware_esxi.add_user user_name=foo password=bar@123 descripton="new user"
+    """
+    log.debug("Running vmware_esxi.add_user")
+    ret = {}
+    if not service_instance:
+        service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
+    hosts = utils_esxi.get_hosts(
+        service_instance=service_instance,
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
+        get_all_hosts=host_name is None,
+    )
+
+    try:
+        for h in hosts:
+            account_spec = vim.host.LocalAccountManager.AccountSpecification()
+            account_spec.id = user_name
+            account_spec.password = password
+            account_spec.description = description
+            h.configManager.accountManager.CreateUser(account_spec)
+            ret[h.name] = True
+        return ret
+    except (
+        vim.fault.InvalidState,
+        vim.fault.NotFound,
+        vim.fault.HostConfigFault,
+        vmodl.fault.InvalidArgument,
+        salt.exceptions.VMwareApiError,
+        vim.fault.AlreadyExists,
+    ) as exc:
+        raise salt.exceptions.SaltException(str(exc))
+
+
+def update_user(
+    user_name,
+    password,
+    description=None,
+    datacenter_name=None,
+    cluster_name=None,
+    host_name=None,
+    service_instance=None,
+):
+    """
+    Update local user on matching ESXi hosts.
+
+    user_name
+        Existing user to update on matching ESXi hosts. (required).
+
+    password
+        New Password for the existing user. (required).
+
+    description
+        New description for the existing user. (optional).
+
+    datacenter_name
+        Filter by this datacenter name (required when cluster is specified)
+
+    cluster_name
+        Filter by this cluster name (optional)
+
+    host_name
+        Filter by this ESXi hostname (optional)
+
+    service_instance
+        Use this vCenter service connection instance instead of creating a new one. (optional).
+
+    .. code-block:: bash
+
+        salt '*' vmware_esxi.update_user user_name=foo password=bar@123 descripton="existing user"
+    """
+    log.debug("Running vmware_esxi.update_user")
+    ret = {}
+    if not service_instance:
+        service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
+    hosts = utils_esxi.get_hosts(
+        service_instance=service_instance,
+        host_names=[host_name] if host_name else None,
+        cluster_name=cluster_name,
+        datacenter_name=datacenter_name,
+        get_all_hosts=host_name is None,
+    )
+
+    try:
+        for h in hosts:
+            account_spec = vim.host.LocalAccountManager.AccountSpecification()
+            account_spec.id = user_name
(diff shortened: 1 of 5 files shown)


src/saltext/vmware/modules/esxi.py
```diff
@@ -845,6 +845,326 @@ def get_ntp_config(
...
+        vim.fault.NotFound,
+        vim.fault.HostConfigFault,
+        vmodl.fault.InvalidArgument,
+        salt.exceptions.VMwareApiError,
+        vim.fault.AlreadyExists,
+    ) as exc:
+        raise salt.exceptions.SaltException(str(exc))
+
+
+def add_user(
```
dhiltonp, line 900: Should `add_user` and `update_user` also accept privileges that should be enabled?

I can do it, is there a reason you didn't add those variables? 

(I'm thinking that maybe it's not done because we don't have code to manage privileges yet?)

src/saltext/vmware/modules/esxi.py
```diff
@@ -845,6 +845,326 @@ def get_ntp_config(
...
+                belongsToGroup=None,
+                belongsToUser=None,
+                domain=None,
+                exactMatch=True,
+                findUsers=True,
+                findGroups=True,
+            )
+            for user in users:
+                ret[h.name] = {
+                    user.principal: {
```
dhiltonp, line 1149: I'm assuming principal is unique, maybe the `username` - is that what it is? It's an odd term.

src/saltext/vmware/modules/esxi.py
```diff
@@ -845,6 +845,326 @@ def get_ntp_config(
...
+                searchStr=user_name,
+                belongsToGroup=None,
+                belongsToUser=None,
+                domain=None,
+                exactMatch=True,
+                findUsers=True,
+                findGroups=True,
+            )
+            for user in users:
+                ret[h.name] = {
```
dhiltonp, line 1148: As written this would only return one user per host, correct - `ret[h.name]` is overwritten for each user.

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,216 @@
...
+        service_instance=service_instance,
+    )
+    hosts = __salt__["vmware_esxi.list_hosts"](
+        datacenter_name=datacenter_name,
+        cluster_name=cluster_name,
+        host_name=host_name,
+        service_instance=service_instance,
+    )
+    for host in hosts:
+        if host in user:
```
dhiltonp, line 88: This line is hard to read

 `user = dict{host: {user}}`, so it works, but even knowing what it was it was unexpected. 

Maybe a name other than `user` would be good for this variable, something like `users_by_host`?

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,216 @@
...
+            # Cannot determine if password is changed. So, when the user is found on the host, update it.
```
dhiltonp, line 89: :'(

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,216 @@
...
+        service_instance=service_instance,
+    )
+    hosts = __salt__["vmware_esxi.list_hosts"](
+        datacenter_name=datacenter_name,
+        cluster_name=cluster_name,
+        host_name=host_name,
+        service_instance=service_instance,
+    )
+    for host in hosts:
+        if host in user:
```
dhiltonp, line 192: same as above, it's just reads like a bug at first glance

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,216 @@
...
+            diff[host] = {
+                "old": {"name": name, "description": user[host][name]["description"]},
+                "new": {"name": name, "description": description},
+                "action": "update",
+            }
+            update += 1
+        else:
+            diff[host] = {"new": {"name": name, "description": description}, "action": "create"}
+            create += 1
+    for host in diff:
```
dhiltonp, line 99: It seems like `update_user` updates multiple hosts, so no iteration is required here.

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,216 @@
...
+        service_instance=service_instance,
+    )
+    for host in hosts:
+        if host in user:
+            delete += 1
+            diff[host] = {name: True}
+        else:
+            diff[host] = {name: False}
+
+    for host in diff:
```
dhiltonp, line 198: Same as above, `remove_user` seems to handle multiple hosts so this loop is unnecessary.

dhiltonp (changes requested):

src/saltext/vmware/modules/esxi.py
```diff
@@ -845,6 +845,326 @@ def get_ntp_config(
         raise salt.exceptions.SaltException(str(exc))
 
 
+def list_hosts(
```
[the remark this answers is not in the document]
cheburakshu: I could have but there are a lot of places where we filter by hosts, but there wasn't any public salt api to list the hosts. So, I used it to solve that problem with this + use it for testing presence.

src/saltext/vmware/modules/esxi.py
cheburakshu: The ticket + module here (https://github.com/ansible-collections/community.vmware/blob/main/plugins/modules/vmware_local_user_manager.py) didn't require me to add them, so just went with the ticket. You can add them if you have a ticket to that effect.

I have another ticket with role that deals with permissions. There is also a group + associating user to a group. But, as Joey said, we are going by priority and dealing with the cluster functionality w/ individual tickets later. (suboptimal but that is how we are doing it now)

src/saltext/vmware/modules/esxi.py
cheburakshu: oops this is a mistake. I have to correct the signature as below as user_name is mandatory in this api. It is supposed to return the user if it exists on the host.

```
def get_user(
    user_name,

```

```
Filter by this user name (required).
```

src/saltext/vmware/modules/esxi.py
cheburakshu: yes, that is the user name. :)

src/saltext/vmware/states/esxi.py
cheburakshu: sure will change

src/saltext/vmware/states/esxi.py
cheburakshu: it is a mixed bag as some of the users need to be created and some others need to be updated on certain hosts. Hence, iterating to create and update users one by one, as appropriate, filtering by host.

src/saltext/vmware/states/esxi.py
cheburakshu: sure will change the name of the var

src/saltext/vmware/states/esxi.py
cheburakshu: loop is necessary again as the remove_user needs to be invoked per host.

dhiltonp: I guess that does make sense. In that case, the `ret` is clobbered on every action; and only the result from the last execution is returned.

src/saltext/vmware/states/esxi.py
dhiltonp: I guess that does make sense. In that case, the `ret` is clobbered on every loop; and only the result from the last execution is returned. That doesn't seem right.

src/saltext/vmware/modules/esxi.py
dhiltonp: I made a mistake, too. 

I was thinking that there might be multiple users per host that match when I read this line, but it doesn't read like that when reading the function.

`exact_match=True` passed to `RetrieveUserGroups` ensures that only one user will exist so my comment was off.

src/saltext/vmware/modules/esxi.py
dhiltonp: I'm confused because reading 3230 I see "I want to add a local role with privileges to an ESXi host".

I didn't see another ticket that deals with privileges, I'm glad you know where it is :)

Now I'm confused between VRAE-3230 and VRAE-3235, one is for roles the other for users? So I guess this PR is for 3235?

src/saltext/vmware/modules/esxi.py
dhiltonp: It might be more clear to use `user_name` instead of `user.principal`, then

src/saltext/vmware/modules/esxi.py
cheburakshu: I am so stupid. I confused you. This PR is for 3235.

cheburakshu: Sorry

src/saltext/vmware/modules/esxi.py
cheburakshu: The value of user.principal which contains the user_name is the key. I am not following your comment. we are not using `user.principal` literal in the dict.

src/saltext/vmware/states/esxi.py
cheburakshu: I thought about it, will think more if there is a way to update it only once after the end of the iteration instead of everytime.

src/saltext/vmware/modules/esxi.py
dhiltonp: We all know you're not stupid, and we all make mistakes :)

Thanks for the clarification!

src/saltext/vmware/modules/esxi.py
dhiltonp: It's not necessary, but if `user_name == user.principal`, then it might be more clear to use `user_name` because it's less obvious what `user.principal` is.

cheburakshu: oh like that ok : 👍

cheburakshu: but it conflicts for others like `user_name.group`, `user_name.id` as opposed to `user.*`. Also user is an object, so user_name wouldn't be a correct usage. Let me know what you think.

cheburakshu: will add comment

src/saltext/vmware/states/esxi.py
cheburakshu: done

dhiltonp: I like the message. 

My only remaining concern is that when running against a large set of hosts, any error will terminate the execution.

Let's say you have 2000 hosts to update but at any time some are down for maintenance. Assuming the list is ordered, you will essentially never be able to update the last machine in the list; if the list were randomized you would eventually be successful.

I think the better approach is to accumulate a list of hosts that had errors, then raise an exception with text like this:

"User {} created on {} host(s) and updated on {} host(s). Errors occurred for {} hosts; sample error: {}"? 

"{} hosts" could be either an integer or a list of hosts (or both)? so: "Errors occurred for {} hosts: {}\n sample error: {}"?

cheburakshu: ok will change

src/saltext/vmware/states/esxi.py
```diff
@@ -1,3 +1,260 @@
...
+                    name, delete, len(failed_hosts), ",".join(failed_hosts), exc
+                )
+                ret["changes"] = diff
+                ret["result"] = False
+            if not ret["comment"]:
+                ret["comment"] = "User {} removed on {} host(s).".format(name, delete)
+                ret["changes"] = diff
+                ret["result"] = True
+        else:
+            if not ret["comment"]:
```
dhiltonp, line 257: I think this should happen at the very end, outside of the loop.  Right now, this message is only emitted if the first host in the list doesn't have the user, and it masks any info about other hosts.

If there were hosts, but there is no comment when we go to return, the user didn't exist on any host.

cheburakshu: ok will do

dhiltonp: Thanks for your patience, Sud :)

cheburakshu:
Thanks to you @dhiltonp for your time and the detailed review.

PR #181, opened 2021-12-10
Tagging

joechainz:


src/saltext/vmware/modules/esxi.py
```diff
@@ -64,6 +64,7 @@ def get_capabilities(service_instance=None):
     if service_instance is None:
         service_instance = get_service_instance(opts=__opts__, pillar=__pillar__)
     hosts = utils_esxi.get_hosts(service_instance=service_instance, get_all_hosts=True)
+    breakpoint()
```
cheburakshu, line 67: may need to be removed

src/saltext/vmware/modules/tag.py
```diff
@@ -0,0 +1,128 @@
...
+__func_alias__ = {"list_": "list"}
+
+
+def __virtual__():
+    if not HAS_PYVMOMI:
+        return False, "Unable to import pyVmomi module."
+    return __virtualname__
+
+
+def create(tag_name, category_id, description=""):
```
cheburakshu, line 29: need state modules + tests.

cheburakshu (changes requested):

src/saltext/vmware/modules/tag.py
```diff
@@ -0,0 +1,131 @@
...
+        spec["update_spec"]["name"] = tag_name
+    if description:
+        spec["update_spec"]["description"] = description
+    url = f"/rest/com/vmware/cis/tagging/tag/id:{tag_id}"
+    response = connect.request(url, "PATCH", body=spec, opts=__opts__, pillar=__pillar__)
+    if response.status_code == 200:
+        return {"tag": "updated"}
+    return {
+        "tag": "failed to update",
+        "status code": response.status_code,
```
cheburakshu, line 89: status_code

src/saltext/vmware/modules/tag.py
```diff
@@ -0,0 +1,131 @@
...
+    tag_id
+        (string) Tag ID of type: com.vmware.cis.tagging.Tag.
+    """
+    url = f"/rest/com/vmware/cis/tagging/tag/id:{tag_id}"
+    response = connect.request(url, "DELETE", opts=__opts__, pillar=__pillar__)
+    if response.status_code == 200:
+        return {"tag": "deleted"}
+    return {
+        "tag": "failed to update",
+        "status code": response.status_code,
```
cheburakshu, line 107: status_code

src/saltext/vmware/states/tag.py
```diff
@@ -0,0 +1,104 @@
...
+__proxyenabled__ = ["vmware_tag"]
+
+
+def __virtual__():
+    if not HAS_PYVMOMI:
+        return False, "Unable to import pyVmomi module."
+    return __virtualname__
+
+
+def manage(name, tag_name=None, tag_description=None, tag_id=None, category_id=None):
```
cheburakshu, line 28: should we have present and absent as two states? manage sounds like an exec module esp since we are requiring the name of the operation from the user.

src/saltext/vmware/states/tag.py
```diff
@@ -0,0 +1,141 @@
...
+                "/rest/com/vmware/cis/tagging/tag",
+                "POST",
+                body=data,
+                token=token,
+                opts=__opts__,
+                pillar=__pillar__,
+            )
+            response = create["response"].json()
+            ret["changes"]["tag_id"] = response["value"]
+            ret["comment"] = "created"
```
cheburakshu, line 98: ret['result'] = True for success

src/saltext/vmware/states/tag.py
```diff
@@ -0,0 +1,141 @@
...
+    for tag in response["value"]:
+        url = f"/rest/com/vmware/cis/tagging/tag/id:{tag}"
+        tag_ref = connect.request(url, "GET", token=token, opts=__opts__, pillar=__pillar__)
+        tag_ref = tag_ref["response"].json()
+        if tag_ref["value"]["name"] == name:
+            found = tag_ref["value"]
+            break
+    if found:
+        if description == found["description"]:
+            ret["comment"] = "tag exists"
```
cheburakshu, line 58: ret['result'] = True for success. We can make the comment more descriptive.

src/saltext/vmware/states/tag.py
```diff
@@ -0,0 +1,141 @@
...
+            url = f"/rest/com/vmware/cis/tagging/tag/id:{id}"
+            updated = connect.request(
+                url, "PATCH", body=spec, token=token, opts=__opts__, pillar=__pillar__
+            )
+            if updated["response"].status_code == 200:
+                ret["changes"]["new"] = {}
+                ret["changes"]["old"] = {}
+                ret["changes"]["old"]["description"] = found["description"]
+                ret["changes"]["new"]["description"] = description
+                ret["comment"] = "updated"
```
cheburakshu, line 72: ret['result'] to be set

src/saltext/vmware/states/tag.py
```diff
@@ -0,0 +1,141 @@
...
+    token = res["token"]
+    found = None
+    for tag in response["value"]:
+        url = f"/rest/com/vmware/cis/tagging/tag/id:{tag}"
+        tag_ref = connect.request(url, "GET", token=token, opts=__opts__, pillar=__pillar__)
+        tag_ref = tag_ref["response"].json()
+        if tag_ref["value"]["name"] == name:
+            found = tag_ref["value"]
+            break
+    if found:
```
cheburakshu, line 56: Need to handle dry run mode (state.apply with test=True).

```
        if __opts__["test"]:
               .....
                ret["comment"] = "tag - {} will be created on resource {}..something like that.."
                ret["result"] = None
       elif ....
```

same for absent.

src/saltext/vmware/states/tag.py
joechainz: result is set in the beginning

cheburakshu: ok, that is good. `ret['result'] = None` for test=True succeeds when you add in that part of the code.

src/saltext/vmware/states/tag.py
joechainz: result is set in the beginning

src/saltext/vmware/states/tag.py
joechainz: result is set in the beginning

src/saltext/vmware/modules/tag.py
```diff
@@ -0,0 +1,131 @@
...
+
+def create(tag_name, category_id, description=""):
+    """
+    Create a new tag.
+
+    tag_name
+        Name of tag.
+
+    category_id
+        (string) Category ID of type: com.vmware.cis.tagging.Tag.
```
waynew, line 37: It would be helpful to provide actual examples of what this even means. I think this is a copy from the API docs but... we can definitely do better here.

tests/integration/modules/test_tag.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_category_list(patch_salt_globals_tag):
+    """
+    Test create tag functionality
+    """
+    cats = tagging.list_category()
+    if len(cats["categories"]) > 0:
+        assert "urn:vmomi:InventoryServiceCategory" in cats["categories"][0]
+    else:
+        pytest.skip("test requires at least one category")
```
waynew, line 15: **question** should we really skip the test, or can we write a fixture that ensures that a category exists?

waynew: Note - the fixture should also just return it, and then this test should depend on the fixture and assert equality.

tests/integration/modules/test_tag.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_tag_list(patch_salt_globals_tag):
+    """
+    Test list tags functionality
+    """
+    tags = tagging.list_()
+    if len(tags["tags"]) > 0:
+        assert "urn:vmomi:InventoryServiceTag:" in tags["tags"][0]
+    else:
+        pytest.skip("test requires at least one tag")
```
waynew, line 26: **suggestion** similar suggestion here - we should go ahead and create the tags that we expect should be there. In fact, I'd probably just write a single `def test_tagging():` test that looks something like this:

```
def test_tagging():
    tag_names = ['test_foo', 'test_bar', 'test_bang']
    for name in tag_names:
        tag = tagging.get(name)
        tagging.delete(tag)  # or whatever the value is
    # make sure no tags are here
    tags = tagging.list_()
    for existing_tag in tags:
        assert not any(existing_tag["name"] == tag for tag in tag_names)
    # make sure we can create a single tag
    tagging.create(name=tag_names[0], ...)
    tag = tagging.get(tag_names[0])
    assert tag["name"] == tag_names[0]
    # create the rest of the tags
     for tag in tag_names[1:]:
        tagging.create(name=tag, ...)
     tags = tagging.list_()
     # now assert all the expected tags exist
     # now update some tags & test them
     # now delete a tag and assert that only it was deleted and the others are still around
    # now delete the rest of the tags and assert that none of them are here
```

Also as I was writing the above, I realized that we should be using the functional test salt minion. I don't know off the top of my head how to create that, but we do that a ton in salt's tests/pytests/functional/ directory.

tests/integration/states/test_tag.py
```diff
@@ -0,0 +1,38 @@
...
+import saltext.vmware.modules.tag as tagging
+import saltext.vmware.states.tag as tagging_state
+
+
+def test_manage_create(patch_salt_globals_tag_state, patch_salt_globals_tag):
+    """
+    test tag manage create
+    """
+    cats = tagging.list_category()
+    if len(cats["categories"]) > 0:
```
waynew, line 13: **suggestion** same suggestion here for the categories/creating fixture

waynew (commented):
I guess it's technically a little late for the review, but... can always open a new PR :sweat_smile: 

One thought I had around the integration tests - could probably just completely do a couple of states using the pytest salt factories minions to create the tags/categories/etc. From a test perspective, as long as the state is calling the modules then we don't need the granular tests.

That being said, I also think we could have a couple of unit tests that check some of the high level bits, e.g. test modes are working, stuff like that.

I can definitely help out on a Test Clinic if needed :+1: 

**praise** I don't actually see anything _wrong_ here, just opportunities to make it excellent! :+1:

PR #185, opened 2021-12-17
Role add/update/remove exec + state mods

cheburakshu:


src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,246 @@ def get_user(
         raise salt.exceptions.SaltException(str(exc))
 
 
+def add_role(
+    role_name,
+    privilege_ids,
+    esxi_host_name=None,
+    esxi_user_name=None,
+    esxi_user_password=None,
```
dhiltonp, line 1174: I'm confused, do the regular service instance credentials not work for setting roles (maybe the role needs to be set after directly connecting to the host)? 

If so, we might need to provide a structured way of having per-host credentials in the pillar data.

src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,246 @@ def get_user(
...
+):
+    """
+    Add local role on ESXi host.
+
+    role_name
+        Role to create on ESXi host. (required).
+
+    privilege_ids
+        List of privileges for the role. (required).
+        Refer: https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.security.doc/GUID-ED56F3C4-77D0-49E3-88B6-B99B8B437B62.html
```
dhiltonp, line 1185: I love the link to documentation right here. 

Looking at the page, I'm not sure what the privileges should look like. 

This is because some names are unusual, like"Alarms.Disable or enable alarm on entity". I wondered if the docs were messed up and the correct name was "Alarms.Disable", but then I also saw "Alarms.Disable alarm action".

A user can figure it out via trial and error, those docs could be improved, or if those *are* correct privileges, I think it would be nice to show them here as an example.

src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,246 @@ def get_user(
...
+            esxi_host=esxi_host_name,
+            esxi_password=esxi_user_password,
+        )
+    try:
+        for role in service_instance.content.authorizationManager.roleList:
+            if role.name == role_name:
+                ret["role_id"] = role.roleId
+                ret["role_name"] = role.name
+                ret["privilege_ids"] = list(role.privilege)
+        return ret
```
dhiltonp, line 1397: If there is no such role, should we raise an exception?

src/saltext/vmware/states/esxi.py
```diff
@@ -25,6 +25,146 @@ def __virtual__():
...
+            ret["result"] = None
+        else:
+            role = __salt__["vmware_esxi.add_role"](
+                role_name=name, privilege_ids=privilege_ids, service_instance=service_instance
+            )
+            changes["new"]["role_id"] = role["role_id"]
+            ret["comment"] = "Role {} created.".format(name)
+            ret["result"] = True
+            ret["changes"] = changes
+    else:
```
dhiltonp, line 97: Maybe `elif new_privs or del_privs:`? Then you can move the "Role {} is in the correct state" be at the end. It keeps this if the same structure as the previous.

This isn't super important, just a thought.

src/saltext/vmware/utils/connect.py
```diff
@@ -36,6 +38,15 @@ def get_service_instance(opts=None, pillar=None):
         will also override any variables of the same name in pillar or
         ext_pillar.
 
+    esxi_host
+        (optioal) If specified, uses this host name instead of other configuration.
```
dhiltonp, line 42: typo: "optional" for these 3 parameters

src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,246 @@ def get_user(
         raise salt.exceptions.SaltException(str(exc))
 
 
+def add_role(
+    role_name,
+    privilege_ids,
```
dhiltonp, line 1171: Why aren't we applying roles to multiple hosts like most other functions?

dhiltonp (changes requested):

src/saltext/vmware/modules/esxi.py
cheburakshu: The AuthorizationManager attribute is available only on the service instance and not at per host level like we did it for users. So, the roles need to be set at every host (pyvmomi shortcoming).

Managing the credentials of every host in the pillar will be a daunting task. We are using call level creds for tags and emulating the same pattern here. Prior discussion here: https://vmware.slack.com/archives/G01PF63FRRS/p1638901094043600

src/saltext/vmware/modules/esxi.py
cheburakshu: That's the best document available. The API has no documentation at all for this enumeration. I can add a few examples, but that will not be exhaustive. Ansible refers to a similar documentation example.

src/saltext/vmware/states/esxi.py
cheburakshu: ok, will change

src/saltext/vmware/utils/connect.py
cheburakshu: will change

src/saltext/vmware/modules/esxi.py
cheburakshu: The AuthorizationManager attribute is available only on the service instance and not at per host level like we did it for users. So, the roles need to be set at every host (pyvmomi shortcoming).

src/saltext/vmware/modules/esxi.py
cheburakshu: don't want to do that. I am using this for checking with a simple `if` condition for state mods and during update. Otherwise need to catch a salt exception to do the same.

src/saltext/vmware/modules/esxi.py
waynew: We may need to have a different discussion about this - I'm suspicious that handling credentials in any way but pillar is a good idea.

cheburakshu: We are currently accepting creds in the calls for all REST APIs today (Tags + NSX). Need to streamline the way things are handled at other places. I will move the handling to pillars for host-wise creds for this PR.

src/saltext/vmware/modules/esxi.py
dhiltonp: I think some documentiation could help clear this up.

"Add local role on ESXi host" makes it sound like it can't apply to the entire vCenter instance.

"Add local role to service instance, which may be an ESXi host or vCenter instance" may be more clear?

cheburakshu: ok will change

src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,210 @@ def get_user(
...
+    esxi_host_name=None,
+    service_instance=None,
+):
+    """
+    Add local role to service instance, which may be an ESXi host or vCenter instance.
+
+    role_name
+        Role to create on ESXi host. (required).
```
dhiltonp, line 1179: maybe remove ESXi host here, or say "on service instance"

src/saltext/vmware/modules/esxi.py
```diff
@@ -1166,6 +1166,210 @@ def get_user(
...
+
+    privilege_ids
+        List of privileges for the role. (required).
+        Refer: https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.security.doc/GUID-ED56F3C4-77D0-49E3-88B6-B99B8B437B62.html
+        Example: ['Folder.Create', 'Folder.Delete'].
+
+    esxi_host_name
```
dhiltonp, line 1186: "Connect to this ESXi host using your pillar's service_instance credentials."?

src/saltext/vmware/modules/esxi.py
cheburakshu: ok will change

src/saltext/vmware/modules/esxi.py
cheburakshu: ok will change

dhiltonp (approved):

PR #188, opened 2022-01-07
Vm snap

joechainz:


src/saltext/vmware/states/vm.py
```diff
@@ -88,3 +88,125 @@ def set_boot_manager(
...
+    if snapshot is None:
+        ret["comment"] = f"Snapshot {snapshot_name} doesn't exist"
+        return ret
+
+    if __opts__["test"]:
+        ret["changes"]["new"] = f"Snapshot {snapshot_name} will be destroyed."
+        ret["comment"] = "These options are set to change."
+        return ret
+
+    utils_vm.destroy_snapshot(snapshot.snapshot, remove_children)
```
waynew, line 208: **question** can this fail? We're never returning a result of False :thinking:

src/saltext/vmware/utils/vm.py
```diff
@@ -789,3 +789,125 @@ def change_boot_options(vm, input_opts):
...
+        current = {
+            "creation_time": str(ss.createTime),
+            "description": ss.description,
+            "id": ss.id,
+            "name": ss.name,
+            "state": ss.state,
+            "quiesced": ss.quiesced,
+        }
+        snaps.append(current)
+        if len(ss.childSnapshotList) > 0:
```
waynew, line 838: **suggestion** prefer idiomatic `if ss.childSnapshotList:` (unless that returns true on an empty one)

src/saltext/vmware/utils/vm.py
```diff
@@ -789,3 +789,125 @@ def change_boot_options(vm, input_opts):
...
+        Name of snapshot to find.
+
+    snapshot_id
+        id of snapshot to find.
+    """
+    for ss in snapshot_tree_root:
+        if snapshot_id is None:
+            if snapshot_name == ss.name:
+                return ss
+            elif len(ss.childSnapshotList) > 0:
```
waynew, line 860: **suggestion** similar suggestion here, and throughout. Rather than `len(...) > 0` just use the boolean nature of empty collections.

tests/integration/modules/test_vm.py
```diff
@@ -39,103 +39,140 @@ def test_vm_list(integration_test_config, patch_salt_globals_vm):
...
-
-def test_powered_on(integration_test_config, patch_salt_globals_vm):
-    """
-    Test deploy virtual machine through an template
+# def test_vm_list_templates(integration_test_config, patch_salt_globals_vm):
+#     """
+#     Test vm list_templates
+#     """
+#     all = virtual_machine.list_templates()
+#     assert all == integration_test_config["virtual_machines_templates"]
```
waynew, line 47: **question** why are these commented out? If they're dead/useless tests, let's remove them. If there's a reason they should be skipped, use `@pytest.skip('reason')` (might be `.mark.skipIf(True, 'reason')`)

waynew (changes requested):
Couple of minor nits - also, `sample.tar` - should that actually be added here?

tests/integration/modules/test_vm.py
joechainz: Oh jeez, this is a full mistake. They were commented so I didn't have to run all of them while I was creating the new tests.

src/saltext/vmware/states/vm.py
joechainz: Yes, it will error if it can not destroy the snap.

waynew: When doing salt states (or modules for that matter) - we shouldn't let exceptions percolate: we should capture them and report something useful to the user.

There's probably a more nuanced view, but generally speaking we should handle errors. For example:

```
local:
----------
          ID: blerp
    Function: neato.kaboom
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/home/wayne/programming/salt/salt/state.py", line 2191, in call
                  ret = self.states[cdata["full"]](
                File "/home/wayne/programming/salt/salt/loader/lazy.py", line 149, in __call__
                  return self.loader.run(run_func, *args, **kwargs)
                File "/home/wayne/programming/salt/salt/loader/lazy.py", line 1201, in run
                  return self._last_context.run(self._run_as, _func_or_method, *args, **kwargs)
                File "/home/wayne/programming/salt/salt/loader/lazy.py", line 1216, in _run_as
                  return _func_or_method(*args, **kwargs)
                File "/home/wayne/programming/salt/salt/loader/lazy.py", line 1249, in wrapper
                  return f(*args, **kwargs)
                File "/home/wayne/salt/var/cache/salt/minion/extmods/states/neato.py", line 16, in kaboom
                  raise Exception('oops')
              Exception: oops
     Started: 17:16:43.886980
    Duration: 1.227 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   1.227 ms
```

That's useful for a developer, but not so much for devops.

```
local:
----------
          ID: blerp
    Function: neato.less_boom
      Result: False
     Comment: Oh dear that was unexpected!
     Started: 17:19:51.625339
    Duration: 0.503 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   0.503 ms
```

would be better.

Basically, the comment should be anything that they need to know to solve the problem.

tests/integration/modules/test_vm.py
waynew: FWIW, next time just use `-k 'some_test'` to run the ones that you're interested in. Or `-k 'not some_other'` to skip the ones you're disinterested in.

Or `@pytest.mark.skipIf(True, 'TODO just testing, bring these back in')`

src/saltext/vmware/states/vm.py
```diff
@@ -88,3 +88,128 @@ def set_boot_manager(
...
+        return ret
+
+    snapshot = utils_vm.create_snapshot(vm_ref, snapshot_name, description, memory, quiesce)
+
+    if isinstance(snapshot, vim.vm.Snapshot):
+        ret["changes"]["new"] = f"Snapshot {snapshot_name} created."
+        ret["comment"] = "created"
+        return ret
+    else:
+        ret["comment"] = "failed to create"
```
waynew, line 153: **suggestion** Should be `ret['result'] = False` here, too, right?

waynew (approved):
:+1: this looks good - we have some other failing integration tests in this module that we need to get sorted out... Perhaps we should plan on beginning of February focusing on getting our tests green.

PR #190, opened 2022-01-12
Ds mm

joechainz:


waynew (changes requested):
One question - otherwise looks good!

src/saltext/vmware/states/datastore.py
```diff
@@ -0,0 +1,88 @@
...
+__proxyenabled__ = ["vmware_datastore"]
+
+
+def __virtual__():
+    if not HAS_PYVMOMI:
+        return False, "Unable to import pyVmomi module."
+    return __virtualname__
+
+
+def maintenance_mode(name, maintenance_mode, datacenter_name=None, service_instance=None):
```
waynew, line 29: **question** would `set_maintenance_mode` be a better name? Also thinking it the argument could then be either renamed to `mode` or `activate_maintenance_mode` :thinking:

waynew: Actually, after a cursory glance at https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.resmgmt.doc/GUID-D229556E-C991-41BD-A378-363CC9D2B2CC.html

I think it makes sense to do:

```suggestion
def maintenance_mode(name, enter_maintenance_mode, datacenter_name=None, service_instance=None):
```

(and update the docstring to match)

src/saltext/vmware/states/datastore.py
```diff
@@ -0,0 +1,88 @@
...
+    service_instance
+        (optional) The Service Instance from which to obtain managed object references.
+    """
+    if service_instance is None:
+        service_instance = connect.get_service_instance(opts=__opts__, pillar=__pillar__)
+    ret = {"name": name, "changes": {}, "result": True, "comment": ""}
+    dc_ref = None
+    if datacenter_name:
+        dc_ref = utils_common.get_datacenter(service_instance, datacenter_name)
+    ds = utils_common.get_datastore(name, dc_ref, service_instance)
```
waynew, line 51: **question** does this correctly work when `dc_ref` is None?

src/saltext/vmware/states/datastore.py
```diff
@@ -0,0 +1,88 @@
...
+    status = ds.summary.maintenanceMode
+    if maintenance_mode:
+        if status == "inMaintenance":
+            ret["comment"] = "Already in maintenance mode."
+            return ret
+
+        if __opts__["test"]:
+            ret["changes"] = {"new": f"datastore {name} will enter maintenance mode."}
+            ret["comment"] = "These options are set to change."
```
waynew, line 60: **suggestion** missing `ret["result"] = None` here for test mode.

src/saltext/vmware/states/datastore.py
```diff
@@ -0,0 +1,88 @@
...
+        ret["result"] = False
+        return ret
+    else:
+        if status == "normal":
+            ret["comment"] = "Already exited maintenance mode."
+            return ret
+
+        if __opts__["test"]:
+            ret["changes"] = {"new": f"datastore {name} will exit maintenance mode."}
+            ret["comment"] = "These options are set to change."
```
waynew, line 78: **suggestion** missing `ret["result"] = None` here for test mode

waynew (changes requested):
Looks good - just missing a couple of bits in the test states

src/saltext/vmware/states/datastore.py
joechainz: yes

waynew (approved):
:rocket: nice!

PR #192, opened 2022-01-13
get datastore info

dhiltonp:
For review, ~~not~~ finished.

Things to do: 
 
 - ~~review `utils/cluster.py:get_cluster` for duplicate functionality~~
 - ~~review `utils/common.py:get_datastore` for duplicate functionality~~
 - ~~review `utils/vmware.py:get_datastores` for duplicate functionality (I'm calling it right now, but as far as I know I just need to take the matching datacenters/clusters/hosts and look at the "datastores" property - I'm calling it in case it does something I don't know about, but I have my doubts?)~~
 - ~~evaluate a few fields/parameters I've left out, see how important they are.~~
 - ~~move `modules/datastore.py:_get_datastores` somewhere more appropriate (depends on the above `get_datastores` exploration)~~
 - ~~make quotes consistent~~


waynew:
Off the top, it also looks like there are several conflicts. Might consider a rebase to resolve those.

src/saltext/vmware/utils/common.py
```diff
@@ -604,6 +604,60 @@ def get_path(node, service_instance, path=""):
...
+    service_instance
+        The Service Instance Object from which to obtain cluster.
+
+    datacenter_name
+        (Optional) Datacenter name to filter by.
+
+    cluster_name
+        (Optional) Exact cluster name to filter by. Requires datacenter_name.
+    """
+    if datacenter_name is None and cluster_name:
```
waynew, line 620: **question** would it be more correct to just check for the truthiness, i.e. `if cluster_name and not datacenter_name:` - in other words, is `''` a valid cluster name?

src/saltext/vmware/utils/common.py
```diff
@@ -604,6 +604,60 @@ def get_path(node, service_instance, path=""):
...
+    if not cluster_name:
+        raise salt.exceptions.ArgumentValueError("Datacenter_name and cluster_name are required when looking up a cluster")
+    clusters = get_clusters(service_instance, datacenter_name, cluster_name)
+    if len(clusters) == 0:
+        return None
+    elif len(clusters) == 1:
+        return clusters[0]
+    else:
+        # this should be unreachable; if hit, it indicates a bug in this function
+        raise Exception()
```
waynew, line 658: **question** would this read better like:

```
if len(clusters) > 1:
    raise WhoopsieDaisy('This should never have happened - please file a bug report at https://github.com/saltstack/salt-ext-modules-vmware/')
else:
    return clusters[0] if clusters else None
```

src/saltext/vmware/utils/common.py
```diff
@@ -729,17 +783,19 @@ def get_vm_datacenter(*, vm):
...
+        if isinstance(mors, type):
+            result = mors
             break
         try:
-            obj = obj.parent
+            mors = mors.parent
         except AttributeError:
             break
-    return datacenter
+    return result
```
waynew, line 801: **suggestion (non-blocking)** this *could* take advantage of early returns, and be simplified with:

```
while True:
    if isinstance(mors, type):
        return mors
    try:
        mors = mors.parent
    except AttributeError:
        return None
```

typically I'm not a fan of early returns, but it actually looks better here :thinking:

tests/integration/conftest.py
```diff
@@ -54,7 +52,7 @@ def salt_call_cli(minion):
 
 @pytest.fixture(scope="session")
 def integration_test_config():
-    default_path = Path().parent.parent / "local" / "vcenter.conf"
+    default_path = Path(__file__).parent.parent.parent / "local" / "vcenter.conf"
```
waynew, line 55: :+1:  That's probably more correct. It will be nice when we're actually setting things up *in* the test suite vs grabbing them.

Actually, it would be worth going ahead and adding a comment here:

> Long term the goal here is that our fixtures should connect to, and configure the vSphere the way we want. This pattern is not one that we really should follow, but at the moment it's the best option that we had.

waynew (changes requested):
Looking good :+1: found a couple of places with some questions, but I didn't catch anything buggy.

src/saltext/vmware/modules/datastore.py
```diff
@@ -0,0 +1,153 @@
...
+    if not objects:
+        return None
+    elif len(objects) == 1:
+        return objects[0]
+    else:
+        # this should be unreacheable and indicates a logic bug in this function
+        # the filters passed to this function should raise an exception if they could be ambiguous,
+        # signaling to the user that their usage is wrong during development, even if run against
+        # an environment with 1 host/cluster/datacenter.
+        raise Exception()
```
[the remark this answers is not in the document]
dhiltonp: I had thought about a NotImplementedError; adding a new exception is also an option.

I don't think it will ever happen, so I'm not sure how much effort to put in to it... but if it does happen it would be nice to get a report!

src/saltext/vmware/utils/common.py
dhiltonp: Fair point, done.

src/saltext/vmware/utils/common.py
dhiltonp: When practicable I do like having the primary flow occur first; the equivalent would then be:

```
    if len(clusters) <= 1:
        return clusters[0] if clusters else None
    else:
        raise Bug('This should never happen - please file a bug report at https://github.com/saltstack/salt-ext-modules-vmware/')
```

I think it does read a little more clearly; I'm still not 100% on having a special "Bug" exception. I'll mull it over.

src/saltext/vmware/utils/common.py
dhiltonp: I agree.

src/saltext/vmware/utils/common.py
dhiltonp: I could also write it:

```
assert len(clusters) <= 1, "please file a bug report at https://github.com/saltstack/salt-ext-modules-vmware/"
return clusters[0] if clusters else None
```

dhiltonp:
Ah, one thing that could be done, but I cannot do now:

`tests/integration/modules/test_datastore.py` and 
`tests/integration/states/test_datastore.py` both use `integration_test_config["datastores"]`.

I have a one liner in `test_value_scraper.py` that populates the value.

However, I'm not pushing it. Instead, we should modify those tests to get a datastore name (cached using a pytest fixture?), then modify it/restore its state.

@joechainz : you're welcome to do that, or I can.

joechainz (approved):

PR #196, opened 2022-01-28
esxi add maintenance mode

cmcmarrow:
esxi add maintenance mode


src/saltext/vmware/modules/esxi.py
```diff
@@ -2059,3 +2059,122 @@ def get(
         return ret
     except DEFAULT_EXCEPTIONS as exc:
         raise salt.exceptions.SaltException(str(exc))
+
+
+def _get_host(host, service_instance=None):
```
joechainz, line 2064: I would put this in utils/esxi.py

joechainz (commented):
Add integration tests

src/saltext/vmware/modules/esxi.py
cmcmarrow: is uses `__opts__` which utils/esxi does not get

waynew:
Ran the pre-commit hooks for you :+1: 

But yeah, this deffo needs tests

joechainz (approved):
Looks good!

PR #197, opened 2022-01-28
One ish click deploy

waynew:
This brings a lot more automation to the release process.

I've updated the documentation to reflect the changes to the release process!

This also adds towncrier for changelog entries, so we need to start requiring changelog entries for almost every change.


joechainz (approved):

waynew:
FWIW we can  @pytest.mark.skip tests (or xfail) tests that are currently broken, if we can't fix them

PR #203, opened 2022-02-07
Skip failing nsxt tests

waynew:
What it says on the tin - these ones are currently failing and must be fixed, but in order to make it easy to require *ALL* tests *MUST* pass before merging, we can skip these for now.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+xfail_strict=true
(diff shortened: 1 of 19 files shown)


waynew:
Also added `xfail_strict = True` -- this will ensure that the tests don't accidentally start working without detection.

To test this, clone the repo and run:  `python -m nox -e tests-3 -- -k nsxt`

Running the full test suite still yields some failures on my machine with a Cava setup, but we've already mentioned needing to get those fixed.

dhiltonp:
`tests/integration/modules/test_nsxt_compute_manager.py:test_get` is failing for me, do you have something configured there that I should also have?

dhiltonp:
I had an old version of `pytest-salt-factories` installed; I've updated and it works for me.

dhiltonp (approved):

PR #205, opened 2022-02-09
Fix several tests

waynew:
There were a couple of places where we were using index access instead of
`.get` or `thing in dict`. This sorted at least several of those out. It also
fixes a bunch of incorrect patching.

We do have some patch code erroneously in the integration tests -- they should
be in the unit tests instead since they're just testing that if a certain error
response happens that an error is raised. There's no need to actually hit a
vCenter for that.

There may be some other integration tests still failing - opening this PR
before they're done running.


waynew:
closing this since the other PR supersedes it

PR #209, opened 2022-02-12
Added vmc-vcenter-stats execution module and corresponding unit tests

shaikres:
Added execution module and corresponding unit tests for vm_vcenter_stats

### DIFF
--- a/src/saltext/vmware/modules/vmc_vcenter_stats.py
+++ b/src/saltext/vmware/modules/vmc_vcenter_stats.py
@@ -70,13 +70,14 @@ def query_monitored_items(
     start_time,
     end_time,
     interval,
-    function,
-    monitored_items_ids,
+    aggregate_function,
+    monitored_items,
     verify_ssl=True,
     cert=None,
 ):
     """
-    Retrieves monitoring data for the given monitored items.
+    Retrieves aggregate monitoring data for the given ``monitored_items`` across the time range.
+    Data will be grouped using the ``aggregate_function`` for each ``interval`` in the time range.
 
     CLI Example:
 
@@ -94,19 +95,19 @@ def query_monitored_items(
        admin password required to login to vCenter console
 
     start_time
-        Start time in UTC. Ex: 2021-05-06T22:13:05.651Z
+        Start time in UTC (inclusive). Ex: 2021-05-06T22:13:05.651Z
 
     end_time
-        End time in UTC. Ex: 2021-05-10T22:13:05.651Z
+        End time in UTC (inclusive). Ex: 2021-05-10T22:13:05.651Z
 
     interval
          interval between the values in hours and mins, for which aggregation will apply.
          Possible values: MINUTES30, HOURS2, MINUTES5, DAY1, HOURS6
 
-    function
+    aggregate_function
         aggregation function. Possible values: COUNT, MAX, AVG, MIN
 
-    monitored_items_ids
+    monitored_items
         List of monitored item IDs. Ex: [cpu.util, mem.util]
 
     verify_ssl
@@ -118,7 +119,7 @@ def query_monitored_items(
         The certificate can be retrieved from browser.
 
     """
-    msg = "Retrieving the vCenter monitoring data for {}".format(monitored_items_ids)
+    msg = "Retrieving the vCenter monitoring data for {}".format(monitored_items)
     log.info(msg)
     api_url_base = vmc_request.set_base_url(hostname)
     api_url = "{base_url}api/appliance/monitoring/query"
@@ -128,9 +129,9 @@ def query_monitored_items(
     params = {
         "start_time": start_time,
         "end_time": end_time,
-        "function": function,
+        "function": aggregate_function,
         "interval": interval,
-        "names": monitored_items_ids,
+        "names": monitored_items,
     }
 
     return vmc_vcenter_request.call_api(
--- a/src/saltext/vmware/modules/vmc_vcenter_stats.py
+++ b/src/saltext/vmware/modules/vmc_vcenter_stats.py
@@ -1,0 +1,144 @@
+"""
+Salt execution module for VMC vCenter stats
+Provides methods to get monitored items list and to query monitored item data for given vCenter.
+To get the data from this module, admin credentials for vCenter are required.
+"""
+import logging
+
+from saltext.vmware.utils import vmc_constants
+from saltext.vmware.utils import vmc_request
+from saltext.vmware.utils import vmc_vcenter_request
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = "vmc_vcenter_stats"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def list_monitored_items(hostname, username, password, verify_ssl=True, cert=None):
+    """
+    Retrieves monitored items list for given vCenter.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt vm_minion vmc_vcenter_stats.list_monitored_items hostname=sample-vcenter.vmwarevmc.com ...
+
+    hostname
+        Hostname of the vCenter console
+
+    username
+        admin username required to login to vCenter console
+
+    password
+       admin password required to login to vCenter console
+
+    verify_ssl
+        (Optional) Option to enable/disable SSL verification. Enabled by default.
+        If set to False, the certificate validation is skipped.
+
+    cert
+        (Optional) Path to the SSL client certificate file to connect to VMC Cloud Console.
+        The certificate can be retrieved from browser.
+
+    """
+
+    log.info("Retrieving the monitored items list for vCenter")
+    api_url_base = vmc_request.set_base_url(hostname)
+    api_url = "{base_url}api/appliance/monitoring"
+    api_url = api_url.format(base_url=api_url_base)
+
+    headers = vmc_vcenter_request.get_headers(hostname, username, password)
+    return vmc_vcenter_request.call_api(
+        method=vmc_constants.GET_REQUEST_METHOD,
+        url=api_url,
+        headers=headers,
+        description="vmc_vcenter_stats.list_monitored_items",
+        verify_ssl=verify_ssl,
+        cert=cert,
+    )
+
+
+def query_monitored_items(
+    hostname,
+    username,
+    password,
+    start_time,
+    end_time,
+    interval,
+    function,
+    monitored_items_ids,
+    verify_ssl=True,
+    cert=None,
+):
+    """
+    Retrieves monitoring data for the given monitored items.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt vm_minion vmc_vcenter_stats.query_monitored_items hostname=sample-vcenter.vmwarevmc.com ...
+
+    hostname
+        Hostname of the vCenter console
+
+    username
+        admin username required to login to vCenter console
+
+    password
+       admin password required to login to vCenter console
+
+    start_time
+        Start time in UTC. Ex: 2021-05-06T22:13:05.651Z
+
+    end_time
+        End time in UTC. Ex: 2021-05-10T22:13:05.651Z
+
+    interval
+         interval between the values in hours and mins, for which aggregation will apply.
+         Possible values: MINUTES30, HOURS2, MINUTES5, DAY1, HOURS6
+
+    function
+        aggregation function. Possible values: COUNT, MAX, AVG, MIN
+
+    monitored_items_ids
+        List of monitored item IDs. Ex: [cpu.util, mem.util]
+
+    verify_ssl
+        (Optional) Option to enable/disable SSL verification. Enabled by default.
+        If set to False, the certificate validation is skipped.
+
+    cert
+        (Optional) Path to the SSL client certificate file to connect to VMC Cloud Console.
+        The certificate can be retrieved from browser.
+
+    """
+    msg = "Retrieving the vCenter monitoring data for {}".format(monitored_items_ids)
+    log.info(msg)
+    api_url_base = vmc_request.set_base_url(hostname)
+    api_url = "{base_url}api/appliance/monitoring/query"
+    api_url = api_url.format(base_url=api_url_base)
+
+    headers = vmc_vcenter_request.get_headers(hostname, username, password)
+    params = {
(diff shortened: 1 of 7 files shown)


waynew (changes requested):
Hey, thanks for the PR! However, there are no integration tests here and when I run against the VMC/SDDC info that I have it just... doesn't return after several seconds.

**suggestion** with an integration test to verify this really does work against VMC, we should be able to merge this.

waynew (changes requested):
I'm still seeing test failures:

```
ERROR tests/integration/modules/test_vmc_dhcp_profiles.py::test_update_dhcp_profile_smoke_test - requests.exceptions.HTTPError: 400 Clien
ERROR tests/integration/modules/test_vmc_networks.py::test_update_network_smoke_test - requests.exceptions.HTTPError: 400 Client Error:
FAILED tests/integration/modules/test_vmc_nat_rules.py::test_create_nat_rule_smoke_test - KeyError: 'id'
```

shaikres:
Added integration tests and also updated testscraper to populate required configuration

src/saltext/vmware/modules/vmc_vcenter_stats.py
```diff
@@ -0,0 +1,144 @@
...
+
+    end_time
+        End time in UTC. Ex: 2021-05-10T22:13:05.651Z
+
+    interval
+         interval between the values in hours and mins, for which aggregation will apply.
+         Possible values: MINUTES30, HOURS2, MINUTES5, DAY1, HOURS6
+
+    function
+        aggregation function. Possible values: COUNT, MAX, AVG, MIN
```
waynew, line 107: **suggestion** `aggregate_function` would probably be a better name for this parameter. With an error of `The following arguments are missing: function` it's not very clear what that argument is *for*, because a function can mean literally anything.

An aggregate function, on the other hand, is clearly only for aggregating data.

src/saltext/vmware/modules/vmc_vcenter_stats.py
```diff
@@ -0,0 +1,144 @@
...
+    start_time,
+    end_time,
+    interval,
+    function,
+    monitored_items_ids,
+    verify_ssl=True,
+    cert=None,
+):
+    """
+    Retrieves monitoring data for the given monitored items.
```
waynew, line 79: **suggestion** since aggregate function is required, and there doesn't appear to be a way to collect discrete data, it makes a lot of sense to say something  like:

```
Retrieves aggregate monitoring data for the given ``monitored_items`` across the time range. Data will be grouped using the ``aggregate_function`` for each ``interval`` in the time range..
```

tests/integration/modules/test_vmc_vcenter_stats.py
```diff
@@ -0,0 +1,98 @@
...
+        "verify_ssl": verify_ssl,
+        "cert": cert,
+    }
+    yield data
+
+
+@pytest.fixture
+def request_headers(common_data):
+    return vmc_vcenter_request.get_headers(
+        common_data["hostname"], common_data["username"], common_data["password"]
```
waynew, line 25: **question** why not just do 

```
def request_headers(vmc_vcenter_admin_connect):
   return vmc_vcenter_request.get_headers(
        common_data["hostname"], common_data["username"], common_data["password"],
        )
```        

Giving things a different name doesn't magically make them a different thing. It's still the same data. In fact:

```
def request_headers(vmc_config):
    common_data = vmc_config["vmc_vcenter_admin_connect"]
```

would work exactly the same as this, and eliminate two fixtures.

tests/integration/modules/test_vmc_vcenter_stats.py
```diff
@@ -0,0 +1,98 @@
...
+def test_list_monitored_items_smoke_test(salt_call_cli, list_monitoring, common_data):
+    ret = salt_call_cli.run("vmc_vcenter_stats.list_monitored_items", **common_data)
+    result = ret.json
+    assert result == list_monitoring
+
+
+def test_query_monitored_items_smoke_test(salt_call_cli, query_monitoring, query_monitoring_params):
+    ret = salt_call_cli.run("vmc_vcenter_stats.query_monitored_items", **query_monitoring_params)
+    result_as_json = ret.json
+    assert result_as_json == query_monitoring
```
waynew, line 98: **question** what is the goal of these tests? What would a failure tell us? All I'm seeing here is that we've just implemented the API call in two different places.

I don't know that we gain much from that.

We have unit tests that assert that arguments are being passed as they should be, and that results are being processed as they should.

Basically I think as long as these results return *any* data, instead of producing an error, then they should be considered correct.

The only alternative for integration tests would be if we could configure the VMC vCenter in such a way that we could ensure that the stats were some particular or range of values. So with this, I'd simply send the arguments and ensure that non-error data was returned. Something like:

```
def test_list_monitored_items_smoke_test(salt_call_cli, vmc_vcenter_admin_connect):
    ret = salt_call_cli.run("vmc_vcenter_stats.query_monitored_items, **vmc_vcenter_admin_connect)  # would probably name the fixture vmc_vcenter_admin_connect_kwargs
    assert 'error' not in ret.json  # or whatever key would indicate an error
```

Basically the same story for the query:

```
def test._query_monitored_items_smoke_test(salt_call_cli, vmc_vcenter_admin_connect_kwargs):
    query_kwargs = {
        "start_time": "2022-02-20T22:13:05.651Z",
        "end_time": "2022-02-21T22:13:05.651Z",
        "interval": "HOURS2",
        "function": "COUNT",
        "monitored_items_ids": "cpu.util,mem.util".split(","),
    }
    ret = salt_call_cli.run("vmc_vcenter_stats.query_monitored_items", **query_kwargs, **vmc_vcenter_admin_connect_kwargs)
    assert 'error' not in ret.json  # also feel free to make other assertions that make sense
```

waynew (changes requested):
Some changes for parameter names, and test simplification.

But I was able to run the modules and see the output, so :+1: it looks good otherwise.

src/saltext/vmware/modules/vmc_vcenter_stats.py
```diff
@@ -0,0 +1,144 @@
...
+        End time in UTC. Ex: 2021-05-10T22:13:05.651Z
+
+    interval
+         interval between the values in hours and mins, for which aggregation will apply.
+         Possible values: MINUTES30, HOURS2, MINUTES5, DAY1, HOURS6
+
+    function
+        aggregation function. Possible values: COUNT, MAX, AVG, MIN
+
+    monitored_items_ids
```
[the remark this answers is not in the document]
shaikres: renamed this param to 'monitored_items'

src/saltext/vmware/modules/vmc_vcenter_stats.py
shaikres: renamed to 'aggregate_function'

src/saltext/vmware/modules/vmc_vcenter_stats.py
shaikres: updated

tests/integration/modules/test_vmc_vcenter_stats.py
shaikres: updated

src/saltext/vmware/modules/vmc_vcenter_stats.py
```diff
@@ -0,0 +1,144 @@
...
+        admin username required to login to vCenter console
+
+    password
+       admin password required to login to vCenter console
+
+    start_time
+        Start time in UTC. Ex: 2021-05-06T22:13:05.651Z
+
+    end_time
+        End time in UTC. Ex: 2021-05-10T22:13:05.651Z
```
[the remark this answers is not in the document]
shaikres: start_time and end_time are inclusive. updated in description

tests/integration/modules/test_vmc_vcenter_stats.py
shaikres: simplified the tests as suggested

tests/unit/modules/test_vmc_vcenter_stats.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def test_list_monitored_items_with_url(mock_request_post_api):
+    expected_url = "https://hostname/api/appliance/monitoring"
+    with patch("saltext.vmware.utils.vmc_vcenter_request.call_api", autospec=True) as vmc_call_api:
+        vmc_vcenter_stats.list_monitored_items(
+            hostname="hostname", username="username", password="password", verify_ssl=False
+        )
+    call_kwargs = vmc_call_api.mock_calls[0][-1]
+    assert call_kwargs["url"] == expected_url
```
dhiltonp, line 83: :+1:

tests/unit/modules/test_vmc_vcenter_stats.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def test_list_monitored_items_should_return_api_response(
+    mock_request_post_api, monitoring_list_data
+):
+    assert (
+        vmc_vcenter_stats.list_monitored_items(
+            hostname="hostname", username="username", password="password", verify_ssl=False
+        )
+        == monitoring_list_data
```
dhiltonp, line 72: This could be a sentinel, but I see the value in having "example" good data.

src/saltext/vmware/modules/vmc_vcenter_stats.py
```diff
@@ -0,0 +1,145 @@
...
+
+    """
+    msg = "Retrieving the vCenter monitoring data for {}".format(monitored_items)
+    log.info(msg)
+    api_url_base = vmc_request.set_base_url(hostname)
+    api_url = "{base_url}api/appliance/monitoring/query"
+    api_url = api_url.format(base_url=api_url_base)
+
+    headers = vmc_vcenter_request.get_headers(hostname, username, password)
+    params = {
```
dhiltonp, line 129: If there were branching or any less straightforward code I would suggest adding a test that parameters line up correctly, but there isn't so I'm not going to.

tools/test_value_scraper_vmc.py
```diff
@@ -102,6 +102,31 @@ def get_vcenter_server_detail(args):
     return output["vcenter_detail"]
 
 
+def get_vcenter_admin_detail(args):
```
dhiltonp, line 105: Wow!

I'm surprised passwords can be read (and not in a good way).

dhiltonp (changes requested):
In general I like how things look.

The code under test has 0 branches so testing that the functions get any data, then verifying that urls are built correctly is probably fine.

In general I suggest using `autospec=True` for `patch`.

tests/integration/conftest.py
```diff
@@ -391,6 +391,23 @@ def vmc_nsx_connect(vmc_config):
     )
 
 
+@pytest.fixture(scope="session")
+def vmc_vcenter_connect(vmc_config):
```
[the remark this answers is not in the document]
shaikres: updated it to just return `vmc_config["vmc_vcenter_connect"]`

This fixture will be used in tests for other modules which will be part of future PRs

tests/integration/modules/test_vmc_vcenter_stats.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_list_monitored_items_smoke_test(salt_call_cli, vmc_vcenter_admin_connect):
+    ret = salt_call_cli.run("vmc_vcenter_stats.list_monitored_items", **vmc_vcenter_admin_connect)
+    assert ret is not None
+    assert "error" not in ret.json
+
+
+def test_query_monitored_items_smoke_test(salt_call_cli, vmc_vcenter_admin_connect):
+    query_kwargs = {
+        "start_time": "2022-02-20T22:13:05.651Z",
```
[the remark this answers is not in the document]
shaikres: we can give any time range to test the function. Having 0 data for the specified time will not return any error.

tests/unit/modules/test_vmc_vcenter_stats.py
shaikres: okay

src/saltext/vmware/modules/vmc_vcenter_stats.py
shaikres: okay

tools/test_value_scraper_vmc.py
shaikres: yes, we can get the required credentials from the response of vmc_sddc.get_by_id

dhiltonp (approved):

dhiltonp (commented):
To be clear, I don't have the SDDC credentials set up; this should not be merged until I have either finished doing that and running the tests, or @waynew verifies the tests function.

At this point my only issue is that there is no test for failure; getting an exception from `call_api` isn't handled or tested for.

src/saltext/vmware/modules/vmc_vcenter_stats.py
waynew: FWIW, the integration test (should) cover this. I mean, arguably it's possible that the parameters (e.g. start/end times) are shuffled. I'm okay with relying on the integration test + visual review to catch discrepancies here :+1:

tests/integration/conftest.py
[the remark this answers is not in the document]
waynew: FWIW it's a bad idea to add test code that isn't used. Just like it's a bad idea to add regular code that isn't used or tested.

Code should only be added in the PR that needs that code, not before. This time it's not a blocker, but next time it will be.

tests/integration/modules/test_vmc_vcenter_stats.py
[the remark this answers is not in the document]
waynew: It *would* be more valuable as a smoke test to have _some_ time range.

tools/test_value_scraper_vmc.py
waynew: Something that we should probably consider doing is adding an `unmask_passwords` parameter, or something similar, and default in all `ret` structures to replacing the passwords with just `<REDACTED>` or something like that.

FWIW, it *is* useful because a map jinja file can use Salt to get the credentials for other states, but... :shrug:

waynew (approved):
Looks good, thanks for making those changes :+1:

PR #211, opened 2022-02-14
Added get_by_id and delete for public_ip module and unit tests

navalkp:
Signed-off-by: Naval Patel <pnaval@vmware.com>

### DIFF
--- a/docs/ref/states/all.rst
+++ b/docs/ref/states/all.rst
@@ -27,4 +27,5 @@ State Modules
     saltext.vmware.states.nsxt_uplink_profiles
     saltext.vmware.states.tag
     saltext.vmware.states.vm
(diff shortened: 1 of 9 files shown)


waynew (changes requested):
Again, unsure if I can't connect to the SDDC (though, other integration tests still work), but when I run this module it just... waits forever at

```
[DEBUG   ] Starting new HTTPS connection (1): example.com:443
```

(where the hostname is the vcenter_hostname from my local/vmc_config.json file)

**suggestion** this PR should also have integration tests to ensure that they can really connect to VMC

navalkp:
Due to some merge issue already checked in files are appearing as diff. So creating a new PR. Closing this one

PR #219, opened 2022-02-21
Fix for vmc test failures - Issue - 218

shaikres:
Fix for test failures


shaikres:
Hi @waynew,  we corrected test set-up and updated all tests. 

Verified with command: `python3 -m nox -e tests-3 -- -k vmc` 


All tests passed in our environment.
image
image


Note: Docs build is failing because of NSX-T modules.

PR #221, opened 2022-02-23
Added Get org users function with unit test coverage

jain-prerna:


src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,106 @@
...
+log = logging.getLogger(__name__)
+
+__virtualname__ = "vmc_org_users"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def get(
```
waynew, line 20: **suggestion** should be `list`

waynew (changes requested):
Just get vs list - integration tests probably aren't super useful without a `create`.

Or it might be OK to just have an integration test that runs and simply asserts than an error isn't returned (maybe that there's an admin user? Not sure what our default is)

src/saltext/vmware/modules/vmc_org_users.py
jain-prerna: Hi Wayne, we followed **get** for list and **get_by_id** for particular item in all the modules.
Do let me know if this need to be renamed , we will do througout .

jain-prerna:
> Just get vs list - integration tests probably aren't super useful without a `create`.
> 
> Or it might be OK to just have an integration test that runs and simply asserts than an error isn't returned (maybe that there's an admin user? Not sure what our default is)


I have other functions as well in this modules, like add and remove
So IT with those tests would be meaningful. Once this merged i will raise all those together

src/saltext/vmware/modules/vmc_org_users.py
waynew: Yes - [that should be done throughout](https://github.com/saltstack/salt-ext-modules-vmware/issues/240)

jain-prerna: okay , i will make this change throughout

jain-prerna: @waynew just pushed the change

waynew (changes requested):
I've updated my vmc_config.json by running `python tools/test_value_scraper_vmc.py -s <the new SDDC ID>` and then run the tests with `python -m nox -e tests-3 -- tests/integration/modules/test_vmc_org_users.py` 

This fails with no fixture `vmc_connect`. After creating the fixture, re-running these tests gives me this:

```
    def test_sddc_smoke_test(salt_call_cli, vmc_common_data):

        # get the org users list

        ret = salt_call_cli.run("vmc_org_users.list", **vmc_common_data)

        result_as_json = ret.json

>       assert "error" not in result_as_json

E       TypeError: argument of type 'NoneType' is not iterable

```

Setting a `breakpoint()` before that line reveals that the ret failed, what looks like because `authorization_host` is being passed to the underlying function. After `pop`ing that out of the config in the test fixture `vmc_common_data`, it was able to list the users, but it failed on the add call.

Changes definitely need to be made here.

waynew:
The `add` function also definitely needs a useful example

tests/integration/modules/test_vmc_org_users.py
```diff
@@ -0,0 +1,46 @@
...
+    # get the org users list
+    ret = salt_call_cli.run("vmc_org_users.list", **vmc_common_data)
+    result_as_json = ret.json
+    assert "error" not in result_as_json
+    existing_org_users = len(result_as_json["results"])
+
+    # Add a new user to the org
+    user_names = ["test@vmware.com"]
+    organization_roles = [
+        {"name": "org_member", "expiresAt": None},
```
waynew, line 29: I get an "illegal request body, illegal value on field: organization roles, illegal value on field: null, illegal value on field: expiresAt" from this

waynew: looks like by removing `expiresAt` from the role it works.

Actually using this during testing, I can't help but think that the UX here could use some attention.

I'd expect something more like:

```
user.add(
    email="test@vmware.com",
    roles=["org_member", "developer"],
)
```

It's entirely possible that I'm missing something with the way the API functions (I took a look at the API docs but they're pretty sparse)

shaikres: we implemented the functionality based on the VMC APIs

shaikres: corrected the tests.

shaikres:
This PR depends on the fixture defined in [PR-251](https://github.com/saltstack/salt-ext-modules-vmware/pull/251)
This needs rebase after the PR-251 is merged.

waynew (changes requested):

tests/integration/modules/test_vmc_org_users.py
```diff
@@ -0,0 +1,54 @@
...
+def vmc_common_data(vmc_connect):
+    data = vmc_connect.copy()
+    data.pop("sddc_id")
+    data.pop("vcenter_hostname")
+    data["hostname"] = vmc_connect["authorization_host"]
+    data.pop("authorization_host")
+    return data
+
+
+def test_sddc_smoke_test(salt_call_cli, vmc_common_data):
```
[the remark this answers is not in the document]
shaikres: Hi @waynew, i have given you the access to perform add user operation. now you will be able to run these tests

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+        refresh_key=refresh_key,
+        authorization_host=hostname,
+        description="vmc_org_users.search",
+        params=params,
+        verify_ssl=verify_ssl,
+        cert=cert,
+    )
+
+
+def add(
```
waynew, line 177: **question** how is this function supposed to work? I'm trying to add a user and it returns successfully, but the user doesn't show up in the `list` or `search` functionality :thinking:

tests/integration/modules/test_vmc_org_users.py
```diff
@@ -0,0 +1,54 @@
...
+    assert "error" not in result_as_json
+    assert result_as_json["results"] == []
+
+    # remove the user from org
+    user_ids = ["vmware.com:test-123"]
+    ret = salt_call_cli.run("vmc_org_users.remove", user_ids=user_ids, **vmc_common_data)
+    result_as_json = ret.json
+    assert "error" not in result_as_json
+    assert result_as_json["succeeded"] == []
+    assert result_as_json["failed"] == user_ids
```
waynew, line 54: **question** why are we deleting a non-existent user? Seems like we would want to be able to test the success scenario.

tests/unit/modules/test_vmc_org_users.py
```diff
@@ -0,0 +1,354 @@
...
+            refresh_key="refresh_key",
+            org_id="org_id",
+            user_search_term="us",
+            verify_ssl=False,
+        )
+        == expected_response
+    )
+
+
+def test_search_sddc_called_with_url():
```
waynew, line 342: **question** search_sddc, eh? :trollface:

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+        The host name of CSP
+
+    refresh_key
+        API Token of the user which is used to get the Access Token required for VMC operations
+
+    org_id
+        The Id of organization for which user list is retrieved
+
+    expand_profile: String
+        (Optional) Indicates if the response should be expanded with the user profile, the value is ignored, only the existence of parameter is checked.
```
waynew, line 54: **suggestion** it's OK if the underlying API doesn't care what the value is. We should. This should be bool with a default of False. Then in the code:

```
if expand_profile:
    prarams['expandProfile'] = 'yes'
```

Or, whatever.

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+
+    include_group_ids_in_roles: String
+        (Optional) Indicates if the inherited roles in the response should indicate group information, the value is ignored, only the existence of parameter is checked.
```
waynew, line 57: **suggestion** same suggestion here

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+    refresh_key
+        API Token of the user which is used to get the Access Token required for VMC operations
+
+    org_id
+        The Id of organization for which user list is retrieved
+
+    user_search_term: String
+        (Required) The string to be searched within email or firstName or lastName or username
+
+    expand_profile: String
```
waynew, line 140: **suggestion** expand_profile should also be bool here

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+                  maxLength: 200
+                  minLength: 0
+                  The resource in which the role is scoped to.
+                  The resource will be embedded in the Access Token "perms" claim, as part of the role
+
+            expiresAt: integer($int64)
+                  example: 3609941597
+                  The timestamp the role expires at (measured in number of seconds since 1/1/1970 UTC)
+
+    service_roles: Array list of ServiceRolesDto
```
waynew, line 253: **suggestion** it's probably good to reference the API terms, but... This doesn't really mean anything. (And I could go on about "Dto")

It'd be something like:

```
service_roles: 
    List of service roles to attach to a user. Example:
    
    .. code::
    
         salt my_minion vmc_org_users.add test@example.com service_roles=[{"serviceDefinitionLink": "some service", "serviceRoles": [{...}, {...}]]
```

Or whatever format those should be in. Basically the users should be able to copy from the docs, make some changes to the particulars, and paste the exact value on the command line.

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+            serviceDefinitionLink: String
+
+            serviceRoles: Array list of ServiceRoleBindingDto
+                  The service roles
+
+    skip_notify_registration: boolean
+        (Optional) Prevent sending mails to users that do not yet have a CSP profile
+
+    invited_by:String
+        (Optional) Invited By, specify the actual user who is inviting
```
waynew, line 265: **question** should this function be `invite` instead of `add`? From what I can tell, it's not actually adding users? :thinking:

src/saltext/vmware/modules/vmc_org_users.py
```diff
@@ -0,0 +1,441 @@
...
+       (Optional) Option to enable/disable SSL verification. Enabled by default.
+       If set to False, the certificate validation is skipped.
+
+    cert
+       (Optional) Path to the SSL client certificate file to connect to VMC Cloud Console.
+       The certificate can be retrieved from browser.
+
+    For example:
+
+              .. code::
```
waynew, line 283: **suggestion** this looks more like some information that would be provided via an HTTP request or Python code, not so much like what would be provided to a module call (I mean, it possibly could be?)

waynew (changes requested):
Okay, I was able to finally make the changes & kick this module around.

The naming of some of the functions is confusing - `add` seems to mean `invite`, in particular. At least, the user that I was adding never showed up in the `list` call or `search`.

There are also a few opportunities to improve the API that we're exposing. And I found one copypasta artifact (there may be others that I missed)

tests/unit/modules/test_vmc_org_users.py
shaikres: corrected

src/saltext/vmware/modules/vmc_org_users.py
shaikres: As we can not expose vmware specific data, we are trying to add a testuser.
The underlying invite API checks if the passed user is part of vmware Active Directory or not. If the user is part of it, then only the underlying invite API will add the user to the given org.

As the testuser is some dummy value and is not part of the vmware Active Directory, the invite API just sends 200 OK response though it did not added the user to the org.

so we are testing the flow as:
- add the user (as the user is not part of vmware Active Directory, the underlying API just sends 200 OK response)
- so when we call search function by  passing the search term as testuser, it should return empty result

tests/integration/modules/test_vmc_org_users.py
shaikres: As mentioned above, we can not use valid vmware user.
Also the underlying remove API will not return error if we pass non-existent user. it will just include that userid in the failed attribute of the response.
 hence  we are just verifying that if we pass non-existent user, the remove API should not return any error and should include the passed userid in the failed attribute of the response.

src/saltext/vmware/modules/vmc_org_users.py
shaikres: changed the function name to invite

src/saltext/vmware/modules/vmc_org_users.py
shaikres: Updated the details as required.

src/saltext/vmware/modules/vmc_org_users.py
shaikres: we are just giving example indicating what value  each field can have.

src/saltext/vmware/modules/vmc_org_users.py
waynew: Sounds like the VMC can be configured, I'm guessing, for different backing LDAP (or other?) account stores. Obviously we don't have any control over that and just have to work within the limits of the API... I think that just calling it `invite` is enough here... even if I'm not happy about it :joy:

tests/integration/modules/test_vmc_org_users.py
waynew: That makes sense, even if I don't like it :upside_down_face:

src/saltext/vmware/modules/vmc_org_users.py
waynew: note that the example isn't actually an example -- that is, I can't copy/paste the cli example and change my hostname & other config values and actually ~create~ invite a user.

src/saltext/vmware/modules/vmc_org_users.py
shaikres: updated properly

src/saltext/vmware/modules/vmc_org_users.py
shaikres: updated properly

src/saltext/vmware/modules/vmc_org_users.py
shaikres: updated properly

src/saltext/vmware/modules/vmc_org_users.py
shaikres: added the proper cli example

PR #222, opened 2022-02-24
Vmc distributed firewall rules create update and corresponding tests 

navalkp:


src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
```diff
@@ -18,6 +18,17 @@ def __virtual__():
...
+    This function creates the payload based on the template and user input passed
+    """
+    data = vmc_request.create_payload_for_request(
+        vmc_templates.create_distributed_firewall_rules, user_input
+    )
+    data["id"] = data["display_name"] = rule_id
+    return data
+
+
 def get(
```
waynew, line 32: **question** should this be `list` instead of get?

src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
```diff
@@ -281,3 +292,461 @@ def delete(
...
+        "description": description,
+        "direction": direction,
+        "notes": notes,
+        "tag": tag,
+        "tags": tags,
+    }
+
+    req_data = vmc_request._filter_kwargs(
+        allowed_kwargs=allowed_dict.keys(), allow_none=["tags"], **allowed_dict
+    )
```
waynew, line 501: **suggestion (non-blocking)** I haven't gone back to look at `_filter_kwargs` but it *seems* like this could be more like just an entry in `allowed_dict` :thinking:

waynew (changes requested):
There are some merge conflicts that will need to be resolved here in addition to the other comments.

src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
shaikres: changed to list

src/saltext/vmware/modules/vmc_distributed_firewall_rules.py
shaikres: `allowed_dict` contains all the input fields (both mandatory and optional) needed for create API.
`_filter_kwargs` is to filter out the non-none values.

shaikres:
resolved merge conflicts and updated code as per review comments

waynew (approved):
:+1:

PR #227, opened 2022-03-02
Fix patch salt globals

joechainz:


tests/integration/states/test_datastore.py
```diff
@@ -1,9 +1,33 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+from unittest.mock import patch
+from venv import create
```
waynew, line 4: **question** was this something that visual studio added for you? :trollface:

joechainz: very odd

PR #228, opened 2022-03-03
Vmc distributed firewall rule state and tests

navalkp:


src/saltext/vmware/states/vmc_distributed_firewall_rules.py
```diff
@@ -0,0 +1,487 @@
...
+        else:
+            return vmc_state._create_state_response(
+                name=name, comment=distributed_firewall_rule["error"], result=False
+            )
+
+    if __opts__.get("test"):
+        log.info("present is called with test option")
+        return vmc_state._create_state_response(
+            name=name,
+            comment="Distributed firewall rule {} will be {}".format(
```
waynew, line 248: **suggestion** it's more common in salt test modes to say `"would have been"` (because nothing is happening nor will it happen unless they actually run the state in non-test mode)

src/saltext/vmware/states/vmc_distributed_firewall_rules.py
```diff
@@ -0,0 +1,487 @@
...
+
+    org_id
+        The Id of organization to which the SDDC belongs to
+
+    sddc_id
+        The Id of SDDC from which the distributed firewall rule should be deleted
+
+    domain_id
+        The domain_id for which the Distributed firewall rule belongs to. Possible values: default, cgw
+
```
waynew, line 404: **question** where is the security_policy_id docs?

waynew (changes requested):
I haven't done a complete review here yet - I do want to actually run some states through this and make sure that it works as expected.

There are a couple of small bits that need to be addressed so far, though.

src/saltext/vmware/states/vmc_distributed_firewall_rules.py
shaikres: missed to add earlier. now updated it

src/saltext/vmware/states/vmc_distributed_firewall_rules.py
shaikres: updated to `would have been`

tests/integration/states/test_vmc_distributed_firewall_rules.py
```diff
@@ -0,0 +1,192 @@
...
+import pytest
+import requests
+from saltext.vmware.utils import vmc_request
+
+
+@pytest.fixture
+def common_data(vmc_config):
+    data = vmc_config["vmc_nsx_connect"].copy()
+    data["domain_id"] = "default"
+    data["security_policy_id"] = "default-layer3-section"
```
waynew, line 13: Not a blocker, but we should have the ability to list security policies. Took me a while to figure out that doesn't exist yet :sweat_smile:

waynew (approved):
The tests all pass and (as mentioned, after hunting around a while) I was able to check this out locally and run the states.

One thing we should probably do, though, is change the state return to use dictdiffer. I opened #247 to track that.

Otherwise, this looks reasonable :+1:

waynew:
(note the tests passed locally. We probably need to pin the jinja version in our test suite or something)

Issue #234, opened 2022-03-08
Bad test password


waynew:
E              reason = 'Weak password: based on personal login information. *** passwd: Authentication token manipulation error'

Looks like the password `password = 'Secret@123'` isn't a good enough test one.

Probably should just do something like:

```
✦ ❯ head -c 10 /dev/urandom | base64
GVh3J69oMcJ0tA==
```
to create the fake password

waynew:
The test is test_user_present_absent_dry_run

PR #235, opened 2022-03-08
fix Weak password

joechainz:


waynew (approved):

waynew:
I guess we'll find out if it was other issues now :joy: Also note this fixes #234 (or at least it should)

PR #239, opened 2022-03-15
nsxt_rest execution module and unit test

kdsalvy:


waynew (commented):
**question** What is the advantage of this module over, say the existing Salt http module? https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.http.html

kdsalvy:
Hi @waynew, this module was provided as a nsxt wrapper to the http module, to handle the use cases for which we don't have a module or state code prioritised as phase 1.

PR #243, opened 2022-03-23
Modification of edge cluster

shaswat91:


tests/unit/modules/test_nsxt_edge_cluster.py
```diff
@@ -0,0 +1,320 @@
...
+    "_revision": 0,
+}
+
+error_json = {"error": "The credentials were incorrect or the account specified has been locked."}
+
+
+@patch.object(nsxt_request, "call_api")
+def test_get_using_basic_auth(mock_call_api):
+
+    mock_call_api.return_value = _mock_edge_clusters
```
waynew, line 135: **suggestion** these should be converted to fixtures. At a glance it looks like there should be two fixtures:

```
@pytest.fixture
def mock_success_api_call(mock_call_api):
    mock_call_api.return_value = {...}  # _mock_edge_clusters
    yield mock_call_api

    
@pytest.fixture
def mock_failure_api_call(mock_call_api):
    mock_call_api.return_value = {...}  # error_json
    yield mock_call_api    
```

waynew (changes requested):
I will definitely need to do a more in-depth review, along with building some sample `module.run` states to run against this, but at the very least the unit tests need some upgrades, and integration tests are also a requirement.

PR #244, opened 2022-03-25
Fixed IP Blocks IT

kdsalvy:


waynew:
Still seeing 

```
E       AssertionError: assert 'error' not in {'error': 'Error occurred while calling NSX-T API https://nsxt-hostname/ip/api/v1/pools/ip-blocks/. Please check logs for more details.'}
```

The other test failures should be sorted, so you should be able to merge the latest from the main branch into your PR (or rebase on main).

waynew:
I *think* you should be able to see the project on the right-hand column - once you get this test passing, feel free to move this back to the backlog column :+1:

kdsalvy:
The test is passing for me, can you please use the config shared?

kdsalvy:
Hi @waynew we can modify the way config is loaded. However, we will still need to replace the config file ( https://github.com/saltstack/salt-ext-modules-vmware/blob/main/tests/integration/nsxt_config.json) with actual values. I'm not sure how to fetch something from the environment if the env related values are not provided in the config. 

I will add the instructions to readme file, to add the top level server and credentials. HTH

waynew:
@kdsalvy are you going to make the changes to this PR or should we go ahead and close it?

kdsalvy:
Hi @waynew , do we need any change in the IT? I have verified it is working fine, on adding the configuration. Any other changes can be taken up as separate PR.

PR #246, opened 2022-03-31
Fix tests

waynew:
Updated the build to use 3004.1 instead - also fixed one test to stop leaving vmware.tgz around (did some other refactoring there, too)


waynew:
Looks like the suites are still failing with the jinja version issue. I think that must have been fixed on the latest salt master branch.

waynew:
only two failing tests now... looking at you mac & windows :joy:

PR #251, opened 2022-04-12
Sddc clusters list , primary, create, and delete fucntions with test coverage

jain-prerna:


src/saltext/vmware/modules/vmc_sddc_clusters.py
```diff
@@ -0,0 +1,319 @@
...
+    """
+    Retrieves the primary cluster in provided customer sddc UUID
+
+    Please refer the `VMC Get primary cluster documentation <https://developer.vmware.com/docs/vmc/latest/vmc/api/orgs/org/sddcs/sddc/primarycluster/get/>`_ to get insight of functionality and input parameters
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt <minion-key-id> vmc_sddc_cluster.get_primary_cluster hostname=vmc.vmware.com ...
```
waynew, line 210: **comment**: that's just "minion id"

tests/integration/conftest.py
```diff
@@ -309,6 +309,21 @@ def vmc_nsx_connect(vmc_config):
     )
 
 
+@pytest.fixture(scope="session")
+def vmc_connect(vmc_config):
+    vmc_config = vmc_config["vmc_connect"]
```
waynew, line 314: **question** why isn't this just `return vmc_config["vmc_connect"]`? At a glance it looks like that would remove the need for two of these fixtures.

tests/integration/modules/test_vmc_sddc_clusters.py
```diff
@@ -0,0 +1,101 @@
...
+        # delete SDDC cluster
+        ret = salt_call_cli.run("vmc_sddc_clusters.delete", cluster_id, **common_data)
+        result_as_json = ret.json
+        assert "error" in result_as_json
+        assert (
+            f"Cluster {cluster_id} is currently not in a state where it can be deleted. Please try once the status is READY or FAILED."
+            == result_as_json["error"]
+        )
+    else:
+        assert "Another cluster creation is in progress" in result_as_json["error"][0]
```
waynew, line 95: **question** is it an accurate statement to say that only one cluster can be in creation at a time?

waynew (changes requested):
Mostly I *think* this looks good - currently my laptop is refusing to start so I can't run the integration tests locally to check on that. I also had a couple of questions.

src/saltext/vmware/modules/vmc_sddc_clusters.py
jain-prerna: okay, will change it to  "minion id"

tests/integration/modules/test_vmc_sddc_clusters.py
jain-prerna: this is  the error message coming from VMC APIs, i am just asserting it here

tests/integration/conftest.py
shaikres: updated

shaikres:
Corrected the test cases and rebased with main branch.
@waynew, please use the new SDDC: Salt-Multihost-SDDC-reshma (e5e387a1-216b-4002-a549-59b24126af5a) to run the tests.

waynew (approved):
This is mostly OK, but the module could be significantly improved.

Specifically, as it's not possible to create a cluster while another one is in progress, that would be very useful information.

Specifically this is the current output:

```
"error": "Another cluster creation is in progress. Please retry once it completes (READY or FAILED)."
```

It would be much nicer if it said something like:

```
status: 
    FAILED
comment:
    Another cluster creation is in progress. Please retry once it completes (READY or FAILED).
waiting for:
    sddc_id: aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee
```

But for now it's probably OK.

PR #253, opened 2022-04-18
sddc Integration test  coverage

jain-prerna:


shaikres:
This PR is dependent on the fixture defined as part of [PR-251](https://github.com/saltstack/salt-ext-modules-vmware/pull/251). After PR-251 is merged, this needs rebase.

@waynew, please use the new SDDC: Salt-Multihost-SDDC-reshma (e5e387a1-216b-4002-a549-59b24126af5a) to run the tests.

waynew (approved):
tests are passing when I run it - though it's kind of a pain because the results can vary, especially depending on when you actually run the test (in other words, typically if you run the whole test suite, or even just a single test suite, and then re-run the suite then you get a different failure mode from the server :grimacing: )

PR #292, opened 2022-07-28
set firewall rule configurations

joechainz:


src/saltext/vmware/modules/esxi.py
```diff
@@ -698,6 +698,210 @@ def get_firewall_config(
         raise salt.exceptions.SaltException(str(exc))
 
 
+def get_firewall_config(
```
dmurphy18, line 701: So above on line 628 we have get_firewall_config -> get_firewall_configs, and here we have get_firewall_config.

If this is a typo, needs fixing, but don't think so due to ruleset_name, then this function is poorly named, since a simple typo and dropped 's' and boom !!!,. not what you expect.

src/saltext/vmware/modules/esxi.py
```diff
@@ -698,6 +698,210 @@ def get_firewall_config(
...
+                firewall_config["name"], host_name=host.name, service_instance=service_instance
+            )
+            ret.append(res)
+        return ret
+
+    except DEFAULT_EXCEPTIONS as exc:
+        raise salt.exceptions.SaltException(str(exc))
+
+
+def set_firewall_configs(
```
dmurphy18, line 856: similar compliant to get_firewall_config and get_firewall_configs above. far too easy to mistype or mis-think the function desired - change the function names to have a clearer name which will prevent misuse or mistake.

dmurphy18 (changes requested):

dmurphy18 (approved):

PR #299, opened 2022-09-13
Escape ansi

joechainz:


cheburakshu (approved):

waynew:
I'm pretty sure there should be a better way to handle this. If you're getting ansi escape sequences on a platform that you shouldn't be getting them in we should be able to detect that and tell the underlying tools to stop outputting said sequences, if they aren't already being handled.

waynew:
Just dropping this here - taking a look I don't actually know why the heck the colors are even being displayed. At appears that TERM is being set on your system in some way that even though it's a subprocess and there is no terminal attached, it thinks there is and it's xterm-256color so it starts spewing ansi escape sequences when _it should not_. It also treats the non-existent terminal as having a size so the output that we're looking for gets missed.

As it turns out, we can bypass this by `env TERM=dumb python3 tools/release.py` - but we also hit the column thing that was problematic. Ended out making some changes to the release script to split on newlines to get them removed from the output :+1:  but this PR shouldn't be necessary, going ahead and closing it!

PR #258, opened 2022-04-27
Service instance profile configuration

abhi1693:
This is a draft for adding support for `profiles` for service instances. The changes should not break any existing functionality. However, there could be more changes needed here. Since I'm unaware of most of the code, I stayed away from doing too many changes that may impact the existing code base.

I have removed the `get_username_password` method as it seemed to be used in only 2 places and replaced it with `get_config` that should handle data being passed from either `__opts__`, `__pillar__` or `env`. It also looks for the `esxi_host` name if it is passed to the method.

I ran the test suite using `VMware-vCenter-Server-Appliance-7.0.3.00500-19480866`

I'm not particularly sure about the files under the `states` module. I have test cases failing for `esxi` and `license_mgr` in there and I cannot seem to figure out why. It could be my changes to `conftest.py` are incorrect.

Also, I don't have a `vmc` setup and all test cases failed related to that failed on my system. This is out of scope for me as I have no knowledge of this product.

Fixes #256 

cc: @waynew

PR #307, opened 2022-09-28
Conf fixes

waynew:
I thought I had made these part of #258 but I guess not :joy: 

As a side note, we probably want to write some tests where we can just either list all the modules or something, but that ensures that `get_service_instance` is being called correctly everywhere :thinking:


waynew (commented):
Just double checked this - if there are no objections I'm going to go ahead and merge tomorrow :+1:

PR #289, opened 2022-07-03
Org users state module with test coverage

jain-prerna:
Current PR is dependent on [PR-221](https://github.com/saltstack/salt-ext-modules-vmware/pull/221)

As PR-221 has execution module and corresponding test coverage for org_users
and current PR has state module and corresponding test coverage for org_users

PR #308, opened 2022-10-18
Fix bandit deps

waynew:
See https://github.com/PyCQA/bandit/issues/956 for specific info, but without adding the additional_dependencies, the plugins fail to load for bandit, oops!


waynew:
This will unblock #289

Issue #288, opened 2022-06-23
vmware_vm.info should have allocated resources


abhi1693:
`vmware_vm.info` should return allocated vCPUs, memory, and total storage as part of the basic information

PR #310, opened 2022-10-20
Adds allocated resources to vm info

abhi1693:
Closes: #288


waynew (changes requested):
I'm guessing that we don't have good test coverage here, since this addition doesn't trigger any test failures :sweat_smile: 

@abhi1693 are you comfortable enough yet to write some tests here?

abhi1693:
@joechainz Hi Joe, is it possible to help merge this PR? I'm not good with pytest and currently dont have an environment to write integration tests.

PR #312, opened 2022-10-20
Add distributed portgroup get method

abhi1693:


waynew:
@abhi1693 I went ahead and wrote some tests... um. A *lot* of thrilling mocking required to actually get this to work - but Thurs/Today's test clinics is where I built that up so they should be helpful in understanding how to dig into the pieces required to mock out pyVmomi stuff.

I'm not going to say that #314 was perfect, but... it *does* work, and it's the way I did it. And the tests are meaningful, which is the most important bit.

abhi1693:
I just saw the tests are failing. I don't think I'll be able to look into this before your test clinic today.

waynew:
note to self (and anyone else) 3.6 was EOL'd, we're not supporting Python versions that are EOL - not gonna block this PR for failing there.

PR #313, opened 2022-10-20
Updates host get method to provide more info on the networking

abhi1693:
The current implementation only returned basic nics information. However, in most cases this is not enough to run automation as we need to know how the VM is connected to the physical port of the host. If it uses DVS or standard switch, and what device the host is connected etc.

This PR solves such problems and provides a much detailed information about the following

- vnics
- pnics
- vswitches
- dvs
- portgroups
- cdp info


waynew (changes requested):
@abhi1693 am I missing something? I don't see any changes here except what looks to be just formatting changes :thinking:

abhi1693:
The PR adds the missing information that may be useful as mentioned in the description

abhi1693:
@joechainz Hi Joe, is it possible to help merge this PR? I'm not good with pytest and currently dont have an environment to write integration tests.

PR #321, opened 2022-11-10
CoE changes for drift report

jgangel:
Less option introduced as well in firewall_config for more readable version.
advanced_config has been modified according to show the differences regarding YAML input file.
storage_report has been added as Test only, no changes being made.
drift_report combines all of them into a single report.


src/saltext/vmware/states/esxi.py
```diff
@@ -927,11 +977,20 @@ def advanced_config(
...
+            ret = {"name": name, "result": True,
+                       "comment": config_input["advanced_options"], "changes": changes}
+            return ret
+        else:
+            ret["result"] = None
+            ret["changes"] = {"new": {}}
+            for host in esxi_config_old:
+                ret["changes"]["new"][host] = f"{name} will be set to {value}"
+                ret["changes"]["old"][host] = f"{name} was {esxi_config_old[host][name]}"
+            ret["comment"] = "These options are set to change."
```
waynew, line 993: **comment**: based on our discussion there are couple of different approaches we can make to extend this - what we *may* want to do is detect which approach the user wants based on what the config values look like and then either manage multiple configs, or just the single one. Of course if just managing a single one isn't that useful, we can go ahead and deprecate that old approach.

Also feel free to update/create new modules if the existing ones either have a bad API or they're too slow because we could get all the data in one call vs several.

src/saltext/vmware/states/esxi.py
```diff
@@ -964,6 +1023,8 @@ def firewall_config(
     cluster_name=None,
     host_name=None,
     service_instance=None,
+    profile=None,
+    less=False
```
waynew, line 1027: **suggestion(non-blocking)** based on our discussion, does advanced config need the `less` parameter as well?

waynew: other alternatives - can we get stuff in `__opts__` (e.g. test=True), or potentially add a different output module?

src/saltext/vmware/states/storage_policy.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+import logging
+
+import saltext.vmware.utils.connect as connect
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = "vsphere_storage_policy"
```
waynew, line 8: **suggestion** per discussion we want to name things once - `vsphere_storage_policy.storage_policy` is twice, should be changed to `vsphere_storage_policy.get` or something to that effect. We also probably want this to be a module, not a state, since it's just getting and not enforcing the state.

src/saltext/vmware/states/storage_policy.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def storage_policy(name, storagePolicies):
```
waynew, line 15: **suggestIon** prefer snake_case, i.e. `s/storagePolicies/storage_policies`

src/saltext/vmware/states/storage_policy.py
```diff
@@ -0,0 +1,34 @@
...
+    Checks state of storage policy of each host
+    policies
+        Policy list to set state to
+    """
+
+    res = connect.request(
+        "/api/vcenter/storage/policies", "GET", opts=__opts__, pillar=__pillar__
+    )
+    response = res["response"].json()
+    current_policies = list(map(lambda p: p['name'], response))
```
waynew, line 26: **suggestion** this can just be a list comprehension instead:

```suggestion
    current_policies = [p["name"] for p in response]
```

Should work exactly the same :+1:

tools/test_value_scraper.py
```diff
@@ -57,9 +57,9 @@ def do_it(*, config_file):
     hosts = si.content.rootFolder.childEntity[0].hostFolder.childEntity[0].host
     for host in hosts:
         config["esxi_host_name"] = host.name
-        config["esxi_datastore_disk_names"] = [
-            extent.diskName for datastore in host.datastore for extent in datastore.info.vmfs.extent
-        ]
+#        config["esxi_datastore_disk_names"] = [
+#            extent.diskName for datastore in host.datastore for extent in datastore.info.vmfs.extent
+#        ]
```
waynew, line 62: **suggestion** probably comment this back in :upside_down_face:

waynew (changes requested):
Here are all the things we discussed that need changes - also remember to add pre-commit either `pre-commit install` or using the vscode extension. Also add `black` to vscode :+1:

src/saltext/vmware/states/drift_report.py
```diff
@@ -0,0 +1,32 @@
...
+def report(name, firewall_config, advanced_config, storage_policy):
+    """
+    Creates drift report
+    policies
+        Policy list to set state to
+    """
+    context = {"__opts__": __opts__, "__salt__": __salt__, "__pillar__": __pillar__}
+    firewall_result = global_injector_decorator(context)(esxi.firewall_config)(**firewall_config)["changes"]
+    advanced_result = global_injector_decorator(context)(esxi.advanced_config)(**advanced_config)["changes"]
+    storage_policy_result = global_injector_decorator(context)(state_storage_policy.storage_policy)(**storage_policy)["changes"]
```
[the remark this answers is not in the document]
jgangel: At the end, we can do this job using returners, so this file was removed from PR

src/saltext/vmware/states/esxi.py
jgangel: As suggested in call, we created a new function, and if the new one ends up being more useful we can deprecate old one

src/saltext/vmware/states/esxi.py
jgangel: We tried to add in __opts__ just by using regular keyword cli argument, but only a set of built-in arguments are valid, the one I found closest to verbosity is log_level, but not sure if we can use this one. For the moment, advanced_config uses less parameter as well

src/saltext/vmware/states/storage_policy.py
jgangel: As suggested in call, storae_policy was renamed to vsphere, and from there we can access several state properties, similar to esxi

src/saltext/vmware/states/storage_policy.py
jgangel: Renamed to policies

src/saltext/vmware/states/storage_policy.py
jgangel: Done

tools/test_value_scraper.py
jgangel: Done

src/saltext/vmware/states/vsphere.py
```diff
@@ -0,0 +1,34 @@
...
+
+    res = connect.request(
+        "/api/vcenter/storage/policies", "GET", opts=__opts__, pillar=__pillar__
+    )
+    response = res["response"].json()
+    current_policies = [p["name"] for p in response]
+    changes = []
+    for policy in policies:
+        if policy['policyName'] not in current_policies:
+            changes.append(policy['policyName'])
```
waynew, line 30: **question** was this supposed to actually change anything? Right now as far as I can tell, it's just *getting* data, which would be expected in a module.

waynew (changes requested):
Mostly looking good, a couple of questions/comments.

src/saltext/vmware/returners/merge_changes_esxi.py
```diff
@@ -0,0 +1,66 @@
...
+                changes["esxi"] = {}
+            for host in state["changes"]:
+                if host not in changes["esxi"]:
+                    changes["esxi"][host] = {}
+                changes["esxi"][host][state["__id__"]] = state["changes"][host]
+        else:
+            changes[state["__id__"]] = state["changes"]
+    with salt.utils.files.flopen(opts["filename"], "a") as logfile:
+        salt.utils.json.dump(changes, logfile)
+        logfile.write("\n")
```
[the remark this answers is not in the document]
jgangel: For the moment, both returners are supported, yes, we can look for alternate solutions later on

src/saltext/vmware/states/esxi.py
```diff
@@ -957,13 +965,100 @@ def advanced_config(
...
+
+    profile
+        Profile to use (optional)
+
+    .. code-block:: yaml
+
+        Remove User:
+        vmware_esxi.advanced_configs:
+            - name: Annotations.WelcomeMessage
+            - value: Hello
```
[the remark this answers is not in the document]
jgangel: I added the indentation for line 1002

src/saltext/vmware/states/esxi.py
```diff
@@ -1008,63 +1103,66 @@ def firewall_config(
...
         for i in range(len(value[name])):
             value[name][i] = dict(value[name][i])
-            if "allowed_host" in value[name][i]:
-                value[name][i]["allowed_host"] = dict(value[name][i]["allowed_host"])
+            if "allowed_hosts" in value[name][i]:
+                value[name][i]["allowed_hosts"] = dict(value[name][i]["allowed_hosts"])
+
+    missing_rules = utils_esxi.get_missing_firewall_rules(value[name], hosts)
+
+    if len(missing_rules) > 0:
```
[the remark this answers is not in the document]
jgangel: Done

src/saltext/vmware/states/esxi.py
```diff
@@ -1008,63 +1103,66 @@ def firewall_config(
...
+        messages = [f"{r[0]} ruleset does not exist on esxi server {r[1]}." for r in missing_rules]
```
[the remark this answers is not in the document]
jgangel: Done

src/saltext/vmware/states/esxi.py
```diff
@@ -1008,63 +1103,66 @@ def firewall_config(
...
-        ret["changes"] = {}
         for host in hosts:
             ret["changes"][host.name] = {}
-            for firewall_config in value[name]:
-                ret["changes"][host.name][firewall_config["name"]] = {}
-                for k in firewall_conf:
+            for ruleset in value[name]:
+                rule = ruleset["name"]
+                ret["changes"][host.name][rule] = {}
+                for k in ruleset:
```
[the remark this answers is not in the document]
jgangel: For the moment, leaving it as it is

src/saltext/vmware/modules/storage_policies.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+def __virtual__():
+    if not HAS_PYVMOMI:
+        return False, "Unable to import pyVmomi module."
+    return __virtualname__
+
+def _connect_spbm_(stub, context):
+   hostname = stub.host.split(":")[0]
+   sessionCookie = stub.cookie.split('"')[1]
```
waynew, line 29: **comment (non-blocking)** this may be fine for a follow-up PR but https://docs.python.org/3/library/http.cookies.html#http.cookies.BaseCookie.load exists and, assuming that the cookie is an actual HTTP valid format for the cookie then it's probably more reliable to use Python's https.cookies to load/parse the cookie value(s)

src/saltext/vmware/modules/storage_policies.py
```diff
@@ -0,0 +1,477 @@
...
+    # old_configs holds only the rules that are in the scope of interest (provided in argument config_input)
+    result = []
+    for policy in result_policies:
+        policy_json = {
+            "policyName": policy.name
+        }
+        policy_json['constraints'] = []
+        try:
+            for sub_profile in policy.constraints.subProfiles:
+                policy_constrain = {
```
waynew, line 337: **question** was this a typo? `s/constrain/constraint`

src/saltext/vmware/modules/storage_policies.py
```diff
@@ -0,0 +1,477 @@
...
+                    "name": sub_profile.name
+                }
+                policy_json['constraints'].append(policy_constrain)
```
waynew, line 340: **suggestion (non-blocking)**: it's a bit of a balancing act when to create a variable vs just use a constant. This *could* be:

```suggestion
                policy_json["constraints"].append({"name": sub_profile.name})
````

Typically I'll use a variable if it makes future debugging attempts easier. In this case, there shouldn't be much of a difference, so _I_ would just use the inline approach.

File: src/saltext/vmware/modules/storage_policies.py [hunk 17]
```diff
@@ -0,0 +1,477 @@
...
+
+                for capability in sub_profile.capability:
+                    for constraint in capability.constraint:
+                        for prop in constraint.propertyInstance:
+                            policy_constrain[prop.id] = prop.value
+        except Exception as err:
```
waynew, line 346: **suggestion (non-blocking)** for this PR it's fine to have the broad exception, but we should strive to have as narrow exceptions as possible. For instance, given the below comment, this could be a TypeError or AttributeError likely, or `for sub_profile in policy.constraints.subProfiles or []:` if the subProfiles is None.

waynew (changes requested):
Looks like pre-commit is failing. Other than that everything looks great - there are a couple of non-blocking issues that can be addressed the follow-up PR. Once pre-commit & tests are passing this should be good to merge :+1:

PR #255, opened 2022-04-24
Added virtual machine filtering

abhi1693:
Closes: #254

PR #322, opened 2022-11-16
Feat/vm filter

waynew:
testing


waynew:
can't push to the branch for #255 - python run was failing on 255, trying to see if this solves it?

waynew:
https://github.com/saltstack/salt-ext-modules-vmware/actions/runs/3475216319/jobs/5809221913 what in tarnation :grimacing:

PR #327, opened 2022-11-28
split manage service

joechainz:


changelog/323.added
```diff
@@ -1 +1 @@
-Added vmware_esxi.set_ntp_config module
\ No newline at end of file
+Added vmware_esxi.set_ntp_config module
```
waynew, line 1: **suggestion** this is the wrong description - should be `service_start, service_stop, service_restart, and service_policy` modules. Also, because we're removing `manage_service` we also need to note that in the changelog - `327.removed`. I *think* it's OK to not bother with deprecating the old function, but we should probably start building in a deprecation policy. So basically this has like... two or three changelogs.

waynew: Oh... I see - this was fixing something or another for pre-commit :sweat_smile: 

But the rest of the comment stands.

waynew (changes requested):

waynew (approved):
:rocket:

PR #329, opened 2022-12-08
create ntp config state

joechainz:


src/saltext/vmware/states/esxi.py
```diff
@@ -1114,3 +1114,168 @@ def firewall_config(
...
+        ``True`` indicates that ntpd should be running and ``False`` indicates that it
+        should be stopped.
+
+    ntp_servers
+        A list of servers that should be added to the ESXi host's NTP configuration.
+
+    service_policy
+        The policy to set for the NTP service.
+        .. note::
+            When setting the service policy to ``off`` or ``on``, you *must* quote the
```
waynew, line 1148: Here's your issue -- 

```suggestion

        .. note::
        
            When setting the service policy to ``off`` or ``on``, you *must* quote the
```

I think that should fix it for you @joechainz

src/saltext/vmware/states/esxi.py
```diff
@@ -1114,3 +1114,170 @@ def firewall_config(
...
+    for host in hosts:
+        if ntp_servers and set(ntp_servers) != set(ntp_config[host.name]["ntp_servers"]):
+            ret["changes"][host.name] = {}
+            if not __opts__["test"]:
+                response = __salt__["vmware_esxi.set_ntp_config"](
+                    ntp_servers=ntp_servers, host_name=host.name, service_instance=service_instance
+                )
+                error = response.get("Error")
+                if error:
+                    ret["comment"] = "Error: {}".format(error)
```
waynew, line 1214: **suggestion** shouldn't `ret['result']` be False here?

src/saltext/vmware/states/esxi.py
```diff
@@ -1114,3 +1114,170 @@ def firewall_config(
...
+                    return ret
+            # Set changes dictionary for ntp_servers
+            ret["changes"][host.name].update(
+                {"ntp_servers": {"old": ntp_config[host.name]["ntp_servers"], "new": ntp_servers}}
+            )
+
+        # Configure service_running state
+        ntp_running = True if ntp_service_state[host.name][ntpd]["state"] == "running" else False
```
waynew, line 1222: **suggestion** just use the result of the comparison

```suggestion
        ntp_running = ntp_service_state[host.name][ntpd]["state"] == "running"
```

src/saltext/vmware/states/esxi.py
```diff
@@ -1114,3 +1114,170 @@ def firewall_config(
...
+                {"service_restart": {"old": "", "new": "NTP Daemon Restarted."}}
+            )
+    ret["result"] = True
+    if ret["changes"] == {}:
+        ret["comment"] = "NTP is already in the desired state."
+        return ret
+
+    if __opts__["test"]:
+        ret["result"] = None
+        ret["comment"] = "NTP state will change."
```
waynew, line 1282: **question** is it not straightforward to do something like:

```
ntp_config = ...

if opts['test']:
    ...
    return something

# make changes here
```
?

waynew (changes requested):
Couple of questions/suggestions

waynew (approved):

PR #330, opened 2022-12-15
vCenter Storage Policies

zruzhinov:
Implementation of vCenter Storage Policies - module, state and tests


waynew:
Note: this requires #321 to be merged first (also the non-blocking comments should be addressed here)

PR #331, opened 2022-12-15
Ignore changelog on end-of-file-fixer

waynew:
We really honestly don't care, since towncrier will handle
stripping/adding newlines where necessary. It's totally fine for
changelog entries if they don't end with a single newline.


waynew:
Will let this percolate - if no objections will merge sometime tomorrow (or later)

waynew:
It was later :joy:

PR #332, opened 2022-12-15
CoE initial ESXi and drift_report

jgangel:
This is a replacement for #321 . Now pointing to fixed branch instead of main/test

### DIFF
--- a/changelog/332.removed
+++ b/changelog/332.removed
@@ -1,1 +1,1 @@
-Removed existing esxi advanced_config state in favor of
+Removed existing esxi advanced_config state in favor of advanced_configs
--- a/changelog/332.removed
+++ b/changelog/332.removed
@@ -1,0 +1,1 @@
+Removed existing esxi advanced_config state in favor of
--- a/src/saltext/vmware/states/esxi.py
+++ b/src/saltext/vmware/states/esxi.py
@@ -3,7 +3,7 @@ import functools
 import json
 import logging
 
-import salt
+import salt.exceptions
 import salt.utils.data
 import salt.utils.dictdiffer
 import saltext.vmware.modules.esxi as vmware_esxi
--- a/src/saltext/vmware/states/esxi.py
+++ b/src/saltext/vmware/states/esxi.py
@@ -888,7 +888,6 @@ def advanced_configs(
     host_name=None,
     service_instance=None,
     profile=None,
-    less=False,
 ):
     """
     Set advanced configuration on matching ESXi hosts.
@@ -911,14 +910,11 @@ def advanced_configs(
     profile
         Profile to use (optional)
 
-    less
-        Default False. If this is set to True, only the changed values will be reported as changes.
 
     .. code-block:: yaml
 
         ESXi_advanced_config_example:
           vmware_esxi.advanced_configs:
-            - less: True
             - configs:
                 DCUI.Access: root
                 Net.BlockGuestBPDU: 1
@@ -936,7 +932,7 @@ def advanced_configs(
         service_instance=service_instance,
     )
 
-    changes = {} if less else {"new": {}, "old": {}}
+    changes = {}
 
     for key in configs:
         for host in esxi_config_old:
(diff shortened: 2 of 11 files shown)


src/saltext/vmware/states/esxi.py
```diff
@@ -877,12 +878,13 @@ def lockdown_mode(
...
-    name,
-    value,
+    name=None,
+    value=None,
     datacenter_name=None,
     cluster_name=None,
     host_name=None,
     service_instance=None,
+    profile=None,
 ):
```
waynew, line 888: **suggestion** let's go ahead and remove advanced_config in favor of `advanced_configs`. Just add `332.removed` to the changelog, probably just `Removed advanced_config in favor of advanced_configs.` And we can pretty easily link to documentation from this PR for anyone who might actually need it :+1:

waynew: Not strictly necessary, since we're removing this function, but: `name` will always be provided (in the example in this docstring it will be "Remove User"), so it shouldn't be given a default of `None`. Also, note that in the docstring `name` and `value` are marked `(required)`, so they shouldn't have a default value unless within the body of the function a check happens.

src/saltext/vmware/states/esxi.py
```diff
@@ -957,13 +965,105 @@ def advanced_config(
     return ret
 
 
+def advanced_configs(
+    configs,
+    name="advanced_configs",
```
waynew, line 970: **suggestion** `name` should be the first parameter for states, since it will always be provided by Salt. In terms of the example, `Remove User` will be the name provided here.

src/saltext/vmware/states/esxi.py
```diff
@@ -957,13 +965,105 @@ def advanced_config(
...
+
+    profile
+        Profile to use (optional)
+
+    .. code-block:: yaml
+
+        Remove User:
+          vmware_esxi.advanced_configs:
+            - name: Annotations.WelcomeMessage
+            - value: Hello
```
waynew, line 1004: **suggestion** this example looks like it was copy-pasted from the other state, and doesn't look like it will actually work with this state.

waynew (changes requested):
Some changes that we discussed in our call today :+1:

src/saltext/vmware/states/esxi.py
jgangel: As advanced_config is removed, this changes are valid in the new advanced_configs

src/saltext/vmware/states/esxi.py
jgangel: Changed to match the actual sls

src/saltext/vmware/states/esxi.py
```diff
@@ -905,56 +905,68 @@ def advanced_config(
...
-        Remove User:
+        ESXi_advanced_config_example:
           vmware_esxi.advanced_configs:
-            - name: Annotations.WelcomeMessage
-            - value: Hello
-
+            - less: True
+            - configs:
+              key_1: value_1
+              key_2: value_2
```
waynew, line 921: **suggestion** these should be indented 2 more spaces, otherwise it's invalid Jinja

```suggestion
                key_1: value_1
                key_2: value_2
```

src/saltext/vmware/states/esxi.py
```diff
@@ -905,56 +905,68 @@ def advanced_config(
...
-            ret["changes"]["new"][host] = f"{name} was changed to {config[host][name]}"
-
-    if change:
-        ret["comment"] = "Configurations have successfully been changed."
-    else:
-        ret["comment"] = "Configurations are already in correct state."
-    return ret
+        for name in configs:
+            value = configs[name]
+            if esxi_config_old[host][name] != value:
```
waynew, line 951: **suggestion** if the config key is not present in the old key, this will error. I'm not sure what the correct/expected behavior should be. We can either create an issue to track this fix, or fix it as part of this PR. I used the literal example (key_1: value_1) when testing this, and that's when it blew up for me :sweat_smile:

waynew (changes requested):
I'll be trying to run the integration tests locally - I did find a couple of issues here that should be addressed. For now I'm OK with creating an issue to track handling the error case, assuming that the "happy path" works, but on future PRs we want to make sure that if someone tries to throw invalid data at us that we don't raise exceptions, but helpful error messages (e.g. "That's not an available config key")

waynew:
FWIW rebasing on main should solve that end-of-file fixer silliness

src/saltext/vmware/states/esxi.py
jgangel: I updated the example with real values and handled the case when the key is not present on any host.

jgangel:
> FWIW rebasing on main should solve that end-of-file fixer silliness

Regarding end of line errors, I get them as Im working on Windows, not sure what the autocrlf should be, I have tried all of them. Currently this passes the pre-commit hook locally so can we try running the workflow again?

waynew:
depends on what editor you're using but autocrlf on windows checkout you should probably use `true`. You should probably do a git reset at that point, I think is what you'd want. I can help you get that tweaked tomorrow :+1:

waynew:
I just pushed the changes we made to my fork - once those get merged  in and the returner change that you wanted to make then I think we're :+1: on this PR.

changelog/332.removed
```diff
@@ -0,0 +1 @@
+Removed existing esxi advanced_config state in favor of
```
waynew, line 1: note - add a `advanced_configs` here.

waynew (changes requested):

waynew (approved):

PR #333, opened 2022-12-16
[CoE] vCenter Storage Policies - moodule, state, and tests

zruzhinov:
Hi @waynew 
this PR includes only the code for storage policies + drift utils.
We didn't have good strategy how to push our code from the fork to the upstream (because of merges in the fork). Now we decide to split PRs into smaller PRs containing only the code for particular functionality/module. It's easier for review. Sorry about duplication.

These are my comments:
- https://github.com/saltstack/salt-ext-modules-vmware/pull/321#discussion_r1049934191
  I'm not sure if this is an HTTP object, this should be just the name of the cookie in quotes.
  
- https://github.com/saltstack/salt-ext-modules-vmware/pull/321#discussion_r1049939446
  the typo is fixed

- https://github.com/saltstack/salt-ext-modules-vmware/pull/321#discussion_r1049941985
   Actually this is variable, I'm using it few lines below for collecting constraint properties in it. Maybe the .append() is confusing, sometimes I forget to add vars in the parent object :), Then I use the ref.

- https://github.com/saltstack/salt-ext-modules-vmware/pull/321#discussion_r1049945805
   Fixed, should be AttributeError...

PR #334, opened 2022-12-20
[CoE] Firewall rules drift report - state and tests

zruzhinov:


waynew (changes requested):
There's a conflict for this, and it looks like #333 also contains some of the changes that are present in this PR?

waynew (approved):
:+1: Nice

PR #335, opened 2023-01-03
[Feat] - Get tags of VM from vSphere SDK Client

oijkn:
With the previous code, the tags did not appear when I did the `salt-call vmware_vm.info <minion>` command.

By adding the vSphere SDK Client, the tags are ok in the return of the command.

**Important**: it is mandatory to install the new dependencies.

```sh
pip3 install --upgrade git+https://github.com/vmware/vsphere-automation-sdk-python.git
```

### DIFF
--- a/src/saltext/vmware/modules/vm.py
+++ b/src/saltext/vmware/modules/vm.py
@@ -9,6 +9,8 @@ import saltext.vmware.utils.connect as connect
 import saltext.vmware.utils.datastore as utils_datastore
 import saltext.vmware.utils.vm as utils_vm
 
+from com.vmware.vapi.std_client import DynamicID
+
 log = logging.getLogger(__name__)
 
 try:
@@ -318,9 +320,7 @@ def info(vm_name=None, service_instance=None, profile=None):
         datacenter_ref = utils_common.get_parent_type(vm, vim.Datacenter)
         mac_address = utils_vm.get_mac_address(vm)
         network = utils_vm.get_network(vm)
-        tags = []
-        for tag in vm.tag:
-            tags.append(tag.name)
+        tags = get_tags(vm._moId)
         folder_path = utils_common.get_path(vm, service_instance)
         info[vm.summary.config.name] = {
             "guest_name": vm.summary.config.name,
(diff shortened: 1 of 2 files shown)


waynew:
@oijkn thanks for the PR!

I was under the impression that tags were get-able with our current approach but maybe that's a specific version of vSphere or something.

A couple of things - first, tests are required for all code contributions. Meaningful unit tests are fine, but unit+integration are better. If you need help with writing tests - attend one of the [Test Clinics on Twitch](https://www.twitch.tv/saltprojectoss/schedule), and we'll be happy to help you get started :+1:. We also want a [changelog entry](https://docs.saltstack.com/en/latest/topics/development/changelog.html)

But more importantly, the vSphere SDK is a very finicky dependency. [It has a known issue about pip-installing](https://github.com/vmware/vsphere-automation-sdk-python/issues/38) for a *very* long time. Almost a decade long.

I've seen that dependency cause a number of issues in upstream Salt, which gives me very *very* strong reservations about making it a dependency of our project.

I'm _OK_ with making it an optional dependency that's basically completely isolated. Something to this effect:


```
import utils.vsphere_sdk


...

    tags = utils.vsphere_sdk.get_tags(vm)
```

Then it could do something like this:


```
try:
    from com.vmware.vapi.std_client import DynamicID
    HAS_VSPHERE_SDK = True
except ImportError:
    HAS_VSPHERE_SDK = False


def get_tags(vm):
    if HAS_VSPHERE_SDK:
        try:
            vm_mid = vm._moId
            ...
            tags = [client.tagging.Tag.get(tag).name for tag in tags_list]
        except Exception:  # not sure if this try/except is necessary
            tags = []
    else:
        tags = [tag.name for tag in vm.tag]
    return tags

```

The create_vsphere_client would also need to be similarly protected. Honestly, I wouldn't put the `api_client` in connect - I'd keep it completely isolated in the vsphere_sdk module. That's just because of the number of issues I've seen using the SDK.

If we can't get the tags any other way :shrug: we can go ahead and provide the ability here. But if someone can't install the SDK, or it gives some issues, then we shouldn't be broken -- we should just carry on as normal, though we can log at an info or warning level or something, if that's useful.

vitos877 (approved):

PR #337, opened 2023-01-05
Pass present state

joechainz:


tests/integration/states/test_esxi.py
```diff
@@ -436,3 +436,23 @@ def test_ntp_config(patch_salt_globals, service_instance):
...
     )
+
+
+def test_password_present_dry_run(patch_salt_globals, dry_run, service_instance):
+    ret = esxi.password_present(
+        name="root",
+        password="TheBigTestPass1!",
+        service_instance=service_instance,
+    )
+    assert ret["result"] is None
```
dmurphy18, line 447: Shouldn't this be True, given the code above defaults result to True

PR #342, opened 2023-01-12
Content library only

jgangel:
This includes modules, state and their corresponding unit tests for content libraries


src/saltext/vmware/modules/subscribed_library.py
```diff
@@ -0,0 +1,127 @@
+# Copyright 2021 VMware, Inc.
```
dmurphy18, line 1: 2021-2023

src/saltext/vmware/states/content_library.py
```diff
@@ -0,0 +1,91 @@
+# SPDX-License-Identifier: Apache-2.0
```
dmurphy18, line 1: missing copyright

tests/unit/modules/test_content_library.py
```diff
@@ -0,0 +1,116 @@
+"""
```
dmurphy18, line 1: copyright and license

tests/unit/states/test_content_library.py
```diff
@@ -0,0 +1,155 @@
+"""
```
dmurphy18, line 1: copyright and license

dmurphy18 (changes requested):
quick glance over the files, and noted we need to get the copyright and license statements in there.  This probably needs to be done for all files to be compliant with the Open Source Office.

Issue #338, opened 2023-01-06
[TECH DEBT] Deprecate/remove test_value_scraper.py


waynew:
Just noticed that this wasn't marked as an issue here yet, so... let's go ahead and create this as an issue!

We took on the test_value_scraper.py debt to be able to start writing tests while we were getting up and running.  This is no longer necessary, and we should be able to define the resources that we need within the conftest.py or other test fixtures.

This is to track the effort to replace that - feel free to edit this as necessary.

PR #344, opened 2023-01-12
Convert to use real .sls file for service instance

waynew:
Part of work for #338 

As we do more for #338, we'll want to document creating `local/vcenter.sls` -- though honestly we could basically create a non-test-value-scraper helper script.

We're gonna need to refactor tests to stop using the `integration_test_config`, but... that should be a good thing :+1: 

As this is, it will skip all of those tests, but we're not running those on PRs anyway, so shouldn't be a big deal.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,14 +28,14 @@ def salt_factories_config():
     }
 
 
-@pytest.fixture(scope="package")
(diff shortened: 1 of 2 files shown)


joechainz (approved):
Looks good!

PR #357, opened 2023-01-27
Vsan config

joechainz:


src/saltext/vmware/modules/esxi.py
```diff
@@ -9,6 +9,7 @@
 import saltext.vmware.utils.esxi as utils_esxi
 import saltext.vmware.utils.vmware as utils_vmware
 from salt.defaults import DEFAULT_TARGET_DELIM
+from salt.ext import six
```
dmurphy18, line 12: do we still need six given Py 3 been around a long time now, if Py 2-ism, surely that should be updated ?

dmurphy18 (dismissed):

dmurphy18 (commented):
Need to fix the conflicts

tools/release.py
```diff
@@ -282,6 +283,11 @@ def commit_changlog_entries(*, version):
             )
 
 
+def escape_ansi(*, line):
+    ansi_escape = re.compile(r"(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]")
```
dmurphy18, line 287: unsure about including UTF-8 directly here, is that a secure and best practices thing to do ?

Just remember issues of spoofing with \x80 which is a space and not valid to do, since they should have used \x20. This was recommended practise back at the end of the last century but don't know if recommendations have changed.

dmurphy18 (commented):
UTF-8 direct usage in regex

tools/release.py
joechainz: Good catch, this isn't even used. So how this got merged in when I was resolving conflicts.

dmurphy18 (approved):

PR #362, opened 2023-02-16
Esxi list disks mod

joechainz:


dmurphy18 (approved):
FYI use of single letters in 'for l' and 'for d' understand C and for i or for j but better practice would be to use a more descriptive name, easier to search for when debugging.

PR #366, opened 2023-03-08
Get host datetime

joechainz:


src/saltext/vmware/modules/esxi.py
```diff
@@ -3673,3 +3673,55 @@ def list_diskgroups(
...
+
+
+def get_host_datetime(
+    host_name=None,
+    datacenter_name=None,
+    cluster_name=None,
+    service_instance=None,
+    profile=None,
+):
+    """
```
Ch3LL, line 3685: I'm not sure how versions are being handled in this salt extension, but would it be worth it to add a versionadded here for this new function?

Ch3LL (changes requested):

src/saltext/vmware/modules/esxi.py
joechainz: The problem with that is this products version is based on the date. Should I just put in a TODO and fill it out on the next release?

Ch3LL: How we used to do it previously with Salt when using date based versions is use code names so for example:

we would add `.. versionadded:: <CODENAME>` and then before release time we would do a search and replace across all files to replace the codename to what the actual date based version is. My comment should not be considered a blocker though. You can merge as is if you want.

twangboy (approved):

Ch3LL (approved):

PR #381, opened 2023-06-05
list ssds module

joechainz:


dmurphy18 (commented):
Wondering about that debug message, surely it should be 'list_ssds'

PR #382, opened 2023-06-07
added list non ssds module to vsphere

joechainz:


tests/integration/modules/test_vsphere.py
```diff
@@ -52,3 +52,13 @@ def test_list_ssds(service_instance):
     for host_name in ret:
         for ssd in ret[host_name]:
             assert isinstance(ssd, str)
+
+
+def test_list_non_ssds(service_instance):
+    ret = vsphere.list_ssds(
```
dmurphy18, line 58: didn't the last PR have a way to return SSDs and non-SSDs so

- Why have a separate function to get non_ssds, see  _get_host_disks, justr don't see need for helper function
- shouldn't this be vsphere.list_non_ssds

dmurphy18 (commented):
wondering about duplication

tests/integration/modules/test_vsphere.py
joechainz: Doh, good catch. Thank you!

tests/integration/modules/test_vsphere.py
```diff
@@ -52,3 +52,13 @@ def test_list_ssds(service_instance):
...
+
+
+def test_list_non_ssds(service_instance):
+    ret = vsphere.list_non_ssds(
+        service_instance=service_instance,
+    )
+    assert ret
+    for host_name in ret:
+        for ssd in ret[host_name]:
+            assert isinstance(ssd, str)
```
dmurphy18, line 64: I was thinking about this, is there a way to check the named ssd is a non-ssd and not a real ssd ?

dmurphy18 (commented):
had a thought about we should check more than just a string

tests/integration/modules/test_vsphere.py
```diff
@@ -52,3 +53,18 @@ def test_list_ssds(service_instance):
...
+def test_list_non_ssds(service_instance):
+    pattern = r"^[a-zA-Z0-9]{3}\."
+    ret = vsphere.list_non_ssds(
+        service_instance=service_instance,
+    )
+    assert ret
+    for host_name in ret:
+        for ssd in ret[host_name]:
+            assert isinstance(ssd, str)
+            if re.match(pattern, ssd):
```
dmurphy18, line 67: That assumes that all names in that regex for non ssds, which at some point will probably become untrue.

tests/integration/modules/test_vsphere.py
```diff
@@ -52,3 +53,18 @@ def test_list_ssds(service_instance):
...
+
+
+def test_list_non_ssds(service_instance):
+    pattern = r"^[a-zA-Z0-9]{3}\."
+    ret = vsphere.list_non_ssds(
+        service_instance=service_instance,
+    )
+    assert ret
+    for host_name in ret:
+        for ssd in ret[host_name]:
```
dmurphy18, line 65: I think nonssd might be a better variable name here, gives quicker flow into the mind, otherwise have to do double-take with ssd , wait I wanted non-ssd, oh! wait etc.

dmurphy18 (commented):
Sorry being picky

dmurphy18 (approved):

PR #386, opened 2023-10-17
add config module dependency.

murugesan-raja:


setup.cfg
```diff
@@ -39,9 +39,10 @@ setup_requires =
     setuptools-declarative-requirements
 install_requires =
     salt>=3005
-    pyvmomi==7.0.3
+    pyvmomi==8.0.1.0
     importlib_metadata; python_version < "3.8"
     jinja2>=3.1.0
+    config-modules@https://github.com/saltstack/salt-ext-modules-vmware/raw/ramurugesan-unified-config-poc/src/saltext/vmware/archive/config_modules-0.0.4-py3-none-any.whl
```
dmurphy18, line 45: @murugesan-raja Adding personal locations into what is supposed to be this branch is not recommended best practice.
Also the failure to run any pre-commit on this PR implies, not taking regular steps for salt repositories.
Can contact me and we can go through this over zoom if that would help

dmurphy18 (changes requested):
Need to not use personal spaces and run pre-commit

setup.cfg
murugesan-raja: @dmurphy18 Thank you for helping with pre-commit. It is now green.

murugesan-raja: About URL specifiers. This approach is mentioned in https://pip.pypa.io/en/stable/reference/requirement-specifiers 
Example:
pip @ https://github.com/pypa/pip/archive/22.0.2.zip
requests [security] @ https://github.com/psf/requests/archive/refs/heads/main.zip ; python_version >= "3.11"

PR #392, opened 2023-11-03
Update pre-commit, py 3.10 support and support for vmotion work-in-progress

dmurphy18:
Updated the pre-commit for updates similar to Salt
Added support for Python 3.10 to pre-commit and testing
Added support for vmotion get|enable|disable - work-in-progress, using it to test the pre-commit and py3.10 changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ setup_requires =
     setuptools>=50.3.2
     setuptools-declarative-requirements
 install_requires =
-    salt>=3005
+    salt>=3006
     pyvmomi==7.0.3
     importlib_metadata; python_version < "3.8"
     jinja2>=3.1.0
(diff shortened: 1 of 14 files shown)


dmurphy18:
Note: disabled Sphinx coverage use since 7.2.6 has a divide by zero issue, which had been fixed and should be available when Sphinx v7.3.0 is released

setup.cfg
```diff
@@ -47,7 +48,7 @@ install_requires =
 tests =
   pytest>=6.1.0
   pytest-cov
-  pytest-salt-factories>=0.121.1
+  pytest-salt-factories>=0.912.2
```
s0undt3ch, line 51: It might require quite a few updates but these days the version that should be getting used is the 1.0.0rcN version

dmurphy18: Would update pytest-salt-factories to 1.0.0rcN, but the version available from PyPI is 0.912.2

dmurphy18: updated to 1.0.0rc27

s0undt3ch: You might have other stuff to fix on tests now

garethgreenaway (approved):

s0undt3ch (approved):

PR #394, opened 2023-11-07
Jruta unified config

Jruta12:
Updated with the tests and documentation for execution module

### DIFF
--- a/src/saltext/vmware/modules/esxi.py
+++ b/src/saltext/vmware/modules/esxi.py
@@ -4478,6 +4478,8 @@ def remediate(profile=None, cluster_paths=None, desired_state_spec=None, esx_con
         # Handle exceptions by setting status as false and including exception details
         log.error("Remediation encountered an error: %s", str(e))
         return {"status": False, "details": str(e)}
+    
+    
 def check_compliance(profile=None, cluster_paths=None, desired_state_spec=None, esx_config=None):
     """
     Checks compliance of cluster.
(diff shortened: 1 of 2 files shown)


dmurphy18:
@Jruta12 Please at least have pre-commit valid before requesting review, even for a work-in-progress this should be a bare minimum, will review once pre-commit issues resolved

Jruta12:
Hi @dmurphy18 , can you please review it now

dmurphy18 (commented):
Needs tests

src/saltext/vmware/modules/esxi.py
```diff
@@ -4156,3 +4156,47 @@ def get_reference_schema():
     """
     log.debug("Running vmware_esxi.retrieve_reference_schema")
     return retrieve_reference_schema(Product.ESX)
+
+
+def convert_ordered_dict_to_dict(obj):
```
dmurphy18, line 4161: Still wondering why we need this since an OrderedDict is a dictionary and what is the benefit of doing this ?   see https://realpython.com/python-ordereddict/, so since Python 3.7 a regular dict is ordered too.
Wondering why you care whether OrderedDict or regular dict, that is, what useful purpose does this function provide.

dmurphy18 (changes requested):
OrderedDict vs Dict question needs answering

src/saltext/vmware/modules/esxi.py
Jruta12: @dmurphy18 we need this function because when we are getting the results from config module they are coming in the form of ordered dictionary and if we will not use this function then the output will not be converted to proper format like yaml

dmurphy18: @Jruta12 that is not making sense to me since OrderedDict is basically a Dictionary at this point with a few extra methods, so you are basically dealing with a dict already.  Perhaps you should change the name of the function since on line 4165 you convert the presented object to a list with the elements of the object which is not reflected in the name of the function or a doc string for it. Relying on the name of the function alone would not convey this.

I would also like to see tests for this function, since I have a feeling it might not do all you think it is doing with, a variety of different objects passed to it.

rprataps: @Jruta12 Remove the orderedDict conversion in the SALT execution module. Config module will take care of handling dict or ordereddict. It has nothing to do with YAML format. From the execution module, when you invoke the config module API, we compare the input vs stored (in VCP DB).  I would change the compare method to handle ordered dict and regular dict, so we wont need to do anything on caller side.

dmurphy18: @rprataps Moving it doesn't  answer my Q as to why have ordered dict vs dict in modern Python, also the function is handling lists too, name of function should change to something that reflects it's real purpose.  And that doesn't negate the need for tests for that function either

rprataps: @dmurphy18 My comment was for Ruta that we don't need the conversion. If you just call config module API without any conversion, it would work. We are not moving the conversion from salt to config module. This conversion is not needed at all.
The comparison function in the config module was failing to compare orderedDict and regularDict.I mean OrderedDict object comes with the Keyword 'OrderedDict' and it was failing while comparing dict and ordered dict objects because of that keyword. It is fixed in the config module now.

Jruta12: > @rprataps Moving it doesn't answer my Q as to why have ordered dict vs dict in modern Python, also the function is handling lists too, name of function should change to something that reflects it's real purpose. And that doesn't negate the need for tests for that function either

@dmurphy18 the function's purpose was to handle the dictionary output coming from config module, but now there is no need for that function because the config module is itself doing the conversion and because we have removed the function itself so no need for tests related to it

Jruta12: @dmurphy18 can you please review the PR, the code has been updated and tested with latest changes and there is no need for "convert_ordered_dict_to_dict" function , please refer the above comment from @rprataps for the same

dmurphy18 (approved):

dmurphy18:
@Jruta12 Need to resolve the conflicts

dmurphy18 (changes requested):
Will Approve, once the conflicts are handled

Jruta12:
@dmurphy18 The conflicts has been resolved now, please check

dmurphy18 (approved):

PR #395, opened 2023-11-07
Apply configuration state module

jgangel:


dmurphy18 (changes requested):
Should NOT be be using 'draft_' as a pre-pend to named functions, a very bad practice.  Using get(key), would be good to provide a default value in case of key not existing , but then you may prefer to have errors thrown if key doesn't exist.

FYI Salt hasn't started using type annotations - too much code and not enough time or people, but if they don't interfere then in favor of leaving them there.

Would expect tests to be developed in conjunction with the code written to test the code as it is being developed, this usually results in better code, see https://www.agilealliance.org/glossary/tdd/ and https://en.wikipedia.org/wiki/Test-driven_development

jgangel (commented):
Draft methods were called like this because this is the naming they currently have in config_modules. These are methods that manipulate drafts within vcenter. For more clarity, though I renamed all methods as [verb]_draft[_complement]

dmurphy18:
:facepalm: Naming, one of the hardest things in computer science :laughing:

dmurphy18 (approved):
with cavets in comments

PR #400, opened 2024-01-29
Minor fix to read config from pillar information and then grain

balavigneshVMware:
For UCM, the assumption is that credentials are passed to pillar first. Read the pillar first and then proceed with other sources, in case of duplicate grains configured with same name as pillar key.


murugesan-raja:
LGTM

PR #407, opened 2024-09-12
Updated pre-commit, GitHub actions, and support of Salt

dmurphy18:
Updated the pre-commit-config.yml to current versions
Updated test.ym for current actions
Resultant files updated for Python 3.10 (currently only supported version of Python in Salt).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -161,10 +161,10 @@ def tests(session):
     args = [
         "--rootdir",
         str(REPO_ROOT),
-        "--log-file={}".format(RUNTESTS_LOGFILE.relative_to(REPO_ROOT)),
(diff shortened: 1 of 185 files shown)


noxfile.py
```diff
@@ -33,7 +34,7 @@
 nox.options.error_on_missing_interpreters = False
 
 # Python versions to test against
-PYTHON_VERSIONS = ("3", "3.7", "3.8", "3.9", "3.10")
+PYTHON_VERSIONS = ("3", "3.9", "3.10")
```
balavigneshVMware, line 37: Given that python support is only from 3.10 `python_requires = >= 3.10`, do we need to test it against 3.9?

dmurphy18: Fixing that, missed a spot

balavigneshVMware (approved):

PR #410, opened 2024-09-16
Modernized the noxfile and added a proto-release github action

dmurphy18:
Testing and updating ability to release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,6 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         print-hash: true
-        skip-existing: true
+        skip-existing: false
         verbose: true
         verify-metadata: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: Release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  Publish:
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.10"
+
+    - name: Install Nox
+      run: |
+        python -m pip install nox
+
+    - name: Build a binary wheel and a source tarball
+      run: |
+        nox -e build
+
+    - name: Publish package distributions to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        print-hash: true
+        verbose: true
+        verify-metadata: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,5 +32,6 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         print-hash: true
+        skip-existing: true
         verbose: true
         verify-metadata: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -418,31 +418,31 @@ jobs:
         path: artifacts/runtests-*.log
 
 
-  Build:
-    runs-on: ubuntu-latest
-    environment: testing
-    permissions:
-      id-token: write
-    needs:
-      - macOS
-      - Windows
-      - Linux
-      - Docs
-
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: "3.10"
-
-    - name: Install Nox
-      run: |
-        python -m pip install nox
-
-    - name: Build a binary wheel and a source tarball
-      run: |
-        nox -e build
+## DGM   Build:
+## DGM     runs-on: ubuntu-latest
+## DGM     environment: testing
+## DGM     permissions:
+## DGM       id-token: write
+## DGM     needs:
+## DGM       - macOS
+## DGM       - Windows
+## DGM       - Linux
+## DGM       - Docs
+## DGM
+## DGM     steps:
+## DGM     - uses: actions/checkout@v4
+## DGM       with:
+## DGM         fetch-depth: 0
+## DGM
+## DGM     - name: Set up Python
+## DGM       uses: actions/setup-python@v5
+## DGM       with:
+## DGM         python-version: "3.10"
+## DGM
+## DGM     - name: Install Nox
+## DGM       run: |
+## DGM         python -m pip install nox
+## DGM
+## DGM     - name: Build a binary wheel and a source tarball
+## DGM       run: |
+## DGM         nox -e build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -447,12 +447,41 @@ jobs:
       run: |
         nox -e build
 
-    - name: Publish distribution 📦 to Test PyPI
-      if: github.event_name.pull_request.merged == true && startsWith(github.ref, 'refs/heads/main')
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -171,6 +171,17 @@ jobs:
...
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        repository_url: https://test.pypi.org/legacy/
+        print-hash: true
+        skip-existing: true
+        verbose: true
+        verify-metadata: true
+
+
```
ScriptAutomate, line 184:
```suggestion
```

Removing for suggested replacement in later suggestion (in order to have a dedicated `Build` step for publishing).

.github/workflows/test.yml
```diff
@@ -295,6 +306,17 @@ jobs:
...
```
ScriptAutomate, line 319:
```suggestion
```

Removing for suggested replacement in later suggestion (in order to have a dedicated `Build` step for publishing).

.github/workflows/test.yml
```diff
@@ -414,3 +436,13 @@ jobs:
...
+    - name: Publish distribution 📦 to Test PyPI
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        repository_url: https://test.pypi.org/legacy/
+        print-hash: true
+        skip-existing: true
+        verbose: true
+        verify-metadata: true
```
ScriptAutomate, line 448:
```suggestion
  Build:
    runs-on: ubuntu-latest
    environment: testing
    permissions:
      id-token: write
    needs:
      - macOS
      - Windows
      - Linux
      - Docs
    steps:
    - uses: actions/checkout@v4
      with:
        fetch-depth: 0
    - name: Set up Python
      uses: actions/setup-python@v5
      with:
        python-version: "3.10"
    - name: Install Nox
      run: |
        python -m pip install nox
    - name: Build a binary wheel and a source tarball
      run: |
        nox -e build

    - name: Publish distribution 📦 to Test PyPI
      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
      uses: pypa/gh-action-pypi-publish@release/v1
      with:
        repository_url: https://test.pypi.org/legacy/
        print-hash: true
        skip-existing: true
        verbose: true
        verify-metadata: true
```

- There should be a separate step doing the build/publish after all other tests have passed, ensuring this only runs once and for one particular version of Python.
- THOUGH: Keep in mind, this is assuming `nox -e build` is how the package is built. You can update my suggestion based on whether a different command should be use.
- My suggestion is also assuming `testing` is the name of the env on GitHub, so can modify the `environment: testing` suggestion if needed
- I'm also using `python-version: "3.10"` as a way of ensuring the uploaded build is done on the version Salt uses in 3006+

ALSO: I think maybe we should evaluate whether it would make sense to do something like a separate workflow that triggers a publish to `test` PYPI whenever a `push` to `main` is done, if we care to have it separate from acting on PRs?

ScriptAutomate (changes requested):

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Build a binary wheel and a source tarball
+      run: |
+        nox -e build
+
+    - name: Publish package distributions to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        print-hash: true
+        skip-existing: true
```
ScriptAutomate, line 35:
```suggestion
        skip-existing: false
```

This should likely only be used when it comes to TestPyPI rather than prod, since test would result in duplicates that can happen but shouldn't be the case with production releases.

ScriptAutomate (changes requested):

.github/workflows/test.yml
dmurphy18: okay will look at doing that tomorrow

.github/workflows/test.yml
dmurphy18: okay will look at doing that tomorrow

.github/workflows/test.yml
dmurphy18: okay will look at doing that tomorrow

.github/workflows/release.yml
dmurphy18: Need to update your doc for suggested release.yml, since that was cut & paste from it
Will fix tomorrow along with the others

.github/workflows/test.yml
dmurphy18: Implemented your suggestion, and researching using ```testing``` but worried about tools other people like NSX in India are using and if doing that would break them.

Do you have any examples of seperate workflows for publish whenever a ```push``` to ``` main``` occurs ?

.github/workflows/release.yml
dmurphy18: done

.github/workflows/test.yml
dmurphy18: done

.github/workflows/test.yml
dmurphy18: done

.github/workflows/test.yml
ScriptAutomate: I think it would look something like this:

```yaml
name: Test PyPi Publish

on:
  push:
    branches:
      - main

jobs:
  Publish:
    runs-on: ubuntu-latest
    environment: testing # Or whatever you name your github environment
    permissions:
      id-token: write
    steps:
    - uses: actions/checkout@v4
      with:
        fetch-depth: 0

    - name: Set up Python
      uses: actions/setup-python@v5
      with:
        python-version: "3.10"

    - name: Install Nox
      run: |
        python -m pip install nox
    - name: Build a binary wheel and a source tarball
      run: |
        nox -e build

    - name: Publish distribution 📦 to Test PyPI
      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
      uses: pypa/gh-action-pypi-publish@release/v1
      with:
        repository_url: https://test.pypi.org/legacy/
        print-hash: true
        skip-existing: true
        verbose: true
        verify-metadata: true
```

Up to us on if we'd want the `if: github.event_name...` conditional check, since if it is present, it would mean it would only trigger the publishing step to Test PyPi if it was a tag. If we keep that check, it would mean we could use a tag that is an `rc` type release version, since that would show whether the publishing to Test PyPi works, helping ensure that a GitHub Release trigger for prod PyPi would have a higher likelihood of known success after.

dmurphy18: Done

.github/workflows/test.yml
```diff
@@ -414,3 +416,72 @@ jobs:
...
+
+          - name: Publish distribution 📦 to Test PyPI
+            if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+            uses: pypa/gh-action-pypi-publish@release/v1
+            with:
+              repository_url: https://test.pypi.org/legacy/
+              print-hash: true
+              skip-existing: true
+              verbose: true
+              verify-metadata: true
```
ScriptAutomate, line 487:
```suggestion
    - name: Publish distribution 📦 to Test PyPI
      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
      uses: pypa/gh-action-pypi-publish@release/v1
      with:
        repository_url: https://test.pypi.org/legacy/
        print-hash: true
        skip-existing: true
        verbose: true
        verify-metadata: true
```

ScriptAutomate (changes requested):

ScriptAutomate (dismissed):

balavigneshVMware (approved):

twangboy (approved):

PR #413, opened 2024-10-09
Only push tagged to test.pypi.org

dmurphy18:


.github/workflows/release.yml
```diff
@@ -2,7 +2,7 @@ name: Release
 
 on:
   release:
-    types: [created]
+    types: [published]
```
ScriptAutomate, line 5:
```suggestion
    types: [released]
```

Ref: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

.github/workflows/release.yml
```diff
@@ -29,6 +29,7 @@ jobs:
         nox -e build
 
     - name: Publish package distributions to PyPI
+      if: github.event.release.released
```
ScriptAutomate, line 32:
```suggestion
```

If `types: [released]` works (from other suggestion), we can remove this if conditional.

.github/workflows/test_publish.yml
```diff
@@ -2,9 +2,8 @@
 name: Test PyPI Publish
 
 on:
-  push:
-    branches:
-      - main
+  release:
+    types: [published]
```
ScriptAutomate, line 6:
```suggestion
    types: [prereleased]
```

Ref: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

.github/workflows/test_publish.yml
```diff
@@ -31,6 +30,7 @@ jobs:
         nox -e build
 
     - name: Publish distribution 📦 to Test PyPI
+      if: github.event.release.prereleased
```
ScriptAutomate, line 33:
```suggestion
```

If `types: [prereleased]` works (from other suggestion), we can remove this if conditional.

ScriptAutomate (changes requested):

ScriptAutomate:
If my suggestions are adopted, it will mean:
- To publish to test pypi: GitHub Releases -> Draft a new release -> create a new tag (with `rc` naming convention) -> check **"Set as a pre-release"** -> publish release
- To publish to prod pypi: GitHub Releases -> Draft a new release -> create a new tag (with `rc` naming convention) -> check **"Set as the latest release"** -> publish release

> **NOTE:** GitHub Releases do NOT do annotated tags. On GitHub, you can see what user did a Release, if needed. You do lose out on some `git` functionality without being able to have the annotated tags, though, with timestamps, `git describe`, user, signing, etc.
>
> Feature request to GitHub: https://github.com/orgs/community/discussions/4924
>
> A workaround would be to figure out how to have the GitHub workflows triggered by prereleases and releases to delete and then recreate the tag but as an annotated tag, if an annotated tag would be desired. That, or have a workflow you can launch that will create an annotated tag, first, before going through the Release page and doing a release based on the new tag. A variant of this could prob be done following https://stackoverflow.com/a/64479344/5340149 (but with a prior step doing `github.rest.git.createTag()` which would be required for an annotated tag before the step doing `github.rest.git.createRef()`.

.github/workflows/release.yml
dmurphy18: done

.github/workflows/test_publish.yml
dmurphy18: done

.github/workflows/release.yml
dmurphy18: Done

.github/workflows/test_publish.yml
dmurphy18: done

ScriptAutomate (approved):

